PR #1, opened 2019-07-26
Mocking dnas

robbiecarlton:
@JettTech @danekszy @lorenjohnson 

Here's my current solution for mocking dna data.

I started by setting up a separate express server to serve a websocket endpoint (see the first commit), but after some discussion with Loren I went with this solution of just switching out the `callZome` as it's lighter weight and less moving parts for testing.

Right now you switch between the real dnas and mock dnas by changing `MOCK_DNA_CONNECTION ` in `holochainClient.js`

It's likely this switch will get more sophisticated later, possibly checking `process.env.NODE_ENV` or another env variable, url components, etc.

The mock data is currently just one big tree, but as it grows we'll likely want to break it into pieces and import them separately

Also I'm experimenting with having functions in the mock data that can update other parts of the tree. (see `register_user`) I can see this being useful for testing more complex interactions between Apollo cache and updating data, eg paginated lists that have items updated in the middle. 

It might turn out that this feature is more trouble than it's worth, we can easily remove it if so.

I'm going to merge this to master to keep things tidy for this initial phase, but don't take that as this being written in stone. Comment on this PR if there's issues you see, or changes you would make.


src/graphql-server/holochainClient.js
```diff
@@ -1,14 +1,17 @@
...
+const MOCK_DNA_CONNECTION = true
 
 export const HOLOCHAIN_LOGGING = true
-const holochainClients = {}
+let holochainClient
 
-async function initAndGetHolochainClient (instanceId) {
-  if (holochainClients[instanceId]) return holochainClients[instanceId]
+async function initAndGetHolochainClient () {
+  if (holochainClient) return holochainClient
```
danekszy, line 12: y u no braces? :(

src/mock-dnas/mockCallZome.js
```diff
@@ -0,0 +1,15 @@
+import mockData from './mockData'
+import { isFunction } from 'lodash/fp'
+
+export default function mockCallZome (instanceId, zome, zomeFunc) {
+  return function (args) {
+    const funcOrResult = mockData[instanceId][zome][zomeFunc]
+    let result
+    if (isFunction(funcOrResult)) {
```
danekszy, line 8: looks like a perfect candidate for ternary ;)

danekszy:
> Also I'm experimenting with having functions in the mock data that can update other parts of the tree.

This is tricky. Usually you don't want to implement state and/or logic to your mocks, since this will leave you with tests testing your mocks' logic, not to mention additional surface for mistakes.
Be careful here.

src/graphql-server/holochainClient.js
robbiecarlton: you don't like those one-liner ifs?

danekszy: Not really. I realise this is probably nitpicking, but on the other hand, there's also [dangling elses](https://en.wikipedia.org/wiki/Dangling_else)

robbiecarlton:
the logic in the mocks is more for when we're developing and manually testing the app, not automated tests. I _think_ we can avoid the situation of automated tests testing the mocks' logic.
And agreed this is tricky. If it turns out to be trouble we'll just turn it off.

danekszy:
Ah, I see. Sounds good.

src/graphql-server/holochainClient.js
robbiecarlton: I don't think I've ever run into a situation where I would want to nest one-liner ifs, I would always break it into multiple lines in that case. The one liner is always (or almost always) in the case where I'm conditionally returning from the function.

PR #2, opened 2019-07-29
(WIP) explore removing dnaInterfaces

robbiecarlton:


danekszy:
Hmmm, I'm ambivalent here. On one hand we drop additional layer of abstraction (which is good), on the other we're losing some visibility into what zomes can be called (documentation sorta).

lorenjohnson:
> Hmmm, I'm ambivalent here. On one hand we drop additional layer of abstraction (which is good), on the other we're losing some visibility into what zomes can be called (documentation sorta).

@robbiecarlton and I talked about this some yesterday and I think the conclusion is that for now we're actually going to keep the abstraction, for the reason you're saying and the fact that the extra abstraction in less REST/CRUD'd like zome interfaces is actually pretty helpful to reducing code repetition where we are composing a set of data from multiple zome calls. 

I can go either way, but generally am on the side of keeping the abstraction. Based upon my experience in getting to this point in the first place, I think it will prove useful in the short term.

danekszy:
Yep, sounds good. The aspect that speaks to me the most is the ability for someone else to look at the code and see exactly what kinds of zome "endpoints" are available. :)

robbiecarlton:
@danekszy - yes, I always felt ambivalent about this change, and as Loren said, ended up deciding to keep the extra layer.

PR #3, opened 2019-07-29
(WIP) As an HP Owner I can browse available hApps

robbiecarlton:


src/graphql-server/index.js
```diff
@@ -2,8 +2,6 @@ import { makeExecutableSchema } from 'graphql-tools'
 import typeDefs from './schema.graphql'
 import resolvers from './resolvers'
 
-console.log('1 - Removing this log statement breaks the app. Investigate why...')
```
danekszy, line 5: Hey @robbiecarlton, did you discover why? :D

danekszy:
One thing I *really* dislike here is how we have HappHosting component in the same directory as RoundImage. This will get messy, super soon. 👎 
/cc @lorenjohnson

lorenjohnson:
If something becomes unmanageable we’ll adjust according to what we find then. Until then let’s stick with components in the component directory and see what comes.

danekszy:
It's not a question of *if*, but *when*, really. We can wait for the trouble and then decide to act on it, but at that point it will become way more expensive for us to untangle.

src/graphql-server/index.js
robbiecarlton: It's something to do with webpack/graphql/caching. The issue is def not resolved, it's just I got it to work without that line. Would be great to pair with someone on the underlying webpack issue.

robbiecarlton:
@danekszy yeah, I also felt uneasy about that. Let's all have a quick chat and try and make a decision about it at end of standup today.

src/graphql-server/index.js
danekszy: Perhaps it has something to do with webpack (or sth else) checking if the module has sideeffects - console is global, so it would probably trigger such check 🤔

PR #4, opened 2019-07-31
(Test Demo 3) Separate component, connector and resolver tests

robbiecarlton:


src/graphql-server/resolvers.test.js
```diff
@@ -0,0 +1,13 @@
+import resolvers from './resolvers'
```
danekszy: It calls the actual zomes, right?

src/components/HappHosting/HappHosting.connector.test.js
```diff
@@ -0,0 +1,56 @@
...
+            hash: 'hash2'
+          }
+        ]
+      }
+    }
+  }
+]
+
+describe('connector', () => {
+  it('runs the AllHapsQuery', async () => {
```
danekszy, line 37: Misleading test name - we're checking the results and not if the query actually executed :P

robbiecarlton: hmm. debatable. Actually all we care about here is that the query ran. So what's misleading is all those `expects`, we probably just want one.

danekszy: Then I would *expect* :P `toHaveBeenCalled` in expect ;)

lorenjohnson (approved):
I prefer this one of the 3 👍

PR #9, opened 2019-08-06
(Do not merge) hha integration code review step 1

robbiecarlton:
@JettTech - thought I would pause here and make some comments while it's still digestable. Tell me if this is helpful and I'll keep doing it.

### DIFF
--- a/src/components/HappHosting/HappHosting.connector.js
+++ b/src/components/HappHosting/HappHosting.connector.js
@@ -7,7 +7,6 @@ import RegisterHostingUserMutation from 'graphql/RegisterHostingUserMutation.gql
 import EnableHappMutation from 'graphql/EnableHappMutation.gql'
 import DisableHappMutation from 'graphql/DisableHappMutation.gql'
 
-
 const allHapps = graphql(AllHappsQuery, {
   props: ({ data: { allHapps } }) => ({ allHapps })
 })
@@ -24,20 +23,31 @@ const hostingUser = graphql(HostingUserQuery, {
   props: ({ data: { hostingUser } }) => ({ hostingUser })
 })
 
-const registerHostingUser = graphql( RegisterHostingUserMutation, {
+const registerHostingUser = graphql(RegisterHostingUserMutation, {
   props: ({ mutate }) => {
     return {
-      // NOTE: Currently host_doc is not validated and the content is irrelevant to the dna...
-      registerHostingUser: ({host_doc}) => mutate({
+      // NOTE: Currently hostDoc is not validated and the content is irrelevant to the dna...
+      registerHostingUser: (hostDoc) => mutate({
         variables: {
-          host_doc
+          hostDoc
+        },
+        update: (cache, { data, data: { registerHostingUser } }) => {
+          console.log('data', data)
+          if (registerHostingUser) {
+            cache.writeQuery({
+              query: HostingUserQuery,
+              data: {
+                hostingUser: registerHostingUser
+              }
+            })
+          }
         }
       })
     }
   }
 })
 
-const enableHapp = graphql( EnableHappMutation, {
+const enableHapp = graphql(EnableHappMutation, {
   props: ({ mutate }) => {
     return {
       enableHapp: (app_hash = '') => mutate({
--- a/src/components/HappHosting/HappHosting.connector.js
+++ b/src/components/HappHosting/HappHosting.connector.js
@@ -27,12 +27,11 @@ const registerHostingUser = graphql(RegisterHostingUserMutation, {
   props: ({ mutate }) => {
     return {
       // NOTE: Currently hostDoc is not validated and the content is irrelevant to the dna...
-      registerHostingUser: (hostDoc) => mutate({
+      registerHostingUser: hostDoc => mutate({
         variables: {
           hostDoc
         },
-        update: (cache, { data, data: { registerHostingUser } }) => {
-          console.log('data', data)
+        update: (cache, { data: { registerHostingUser } }) => {
           if (registerHostingUser) {
             cache.writeQuery({
               query: HostingUserQuery,
--- a/src/components/HappHosting/HappHosting.js
+++ b/src/components/HappHosting/HappHosting.js
@@ -5,17 +5,15 @@ import HashIcon from 'components/HashIcon'
 import SpecificButton from 'components/SpecificButton'
 
 export default function HappHosting ({ allHapps, allAvailableHapps, allHostedHapps, registerHostingUser, hostingUser, enableHapp, disableHapp }) {
-
-  console.log("Host WHOAMI Result >> ", hostingUser)
+  console.log('Host WHOAMI Result >> ', hostingUser)
 
   return <div>
-    {hostingUser && hostingUser.result ?
-       <h2>Current Host: {hostingUser.result}</h2>
-    :
-       <SpecificButton styleName='center-items' onClick={()=>registerHostingUser({host_doc:"info to register host"})} >Register as Host</SpecificButton>
-     }
+    {hostingUser && hostingUser.id
+      ? <h2>Current Host: {hostingUser.id}</h2>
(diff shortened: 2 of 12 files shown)


src/components/HappHosting/HappHosting.connector.js
```diff
@@ -24,20 +23,31 @@ const hostingUser = graphql(HostingUserQuery, {
   props: ({ data: { hostingUser } }) => ({ hostingUser })
 })
 
-const registerHostingUser = graphql( RegisterHostingUserMutation, {
+const registerHostingUser = graphql(RegisterHostingUserMutation, {
   props: ({ mutate }) => {
     return {
-      // NOTE: Currently host_doc is not validated and the content is irrelevant to the dna...
-      registerHostingUser: ({host_doc}) => mutate({
+      // NOTE: Currently hostDoc is not validated and the content is irrelevant to the dna...
```
robbiecarlton, line 29: renamed this variable from `host_doc` to `hostDoc` everywhere as Standard JS wants everything to be camel case.

src/components/HappHosting/HappHosting.connector.js
```diff
@@ -24,20 +23,31 @@ const hostingUser = graphql(HostingUserQuery, {
...
+      registerHostingUser: (hostDoc) => mutate({
         variables: {
-          host_doc
+          hostDoc
+        },
+        update: (cache, { data, data: { registerHostingUser } }) => {
```
robbiecarlton: this whole `update` function is the thing I was talking about "the last thing you haven't seen about Apollo". Basically, it's a callback that gets called after the mutation returns. it gets passed the _result_ of the mutation, as well as the cache. The cache is the Apollo equivalent of the redux store, it's where all the local state is stored, specifically the results of all our graphQL queries.

So what I'm doing here is taking the result of the `registerHostingUser` mutation and writing it to the cache under the `HostingUserQuery`, so that when Apollo then goes to fetch the results of `HostingUserQuery`, it will return the result of `registerHostingUser`, and won't have to call the dna again to get the result.

Most often we get this kind of behavior for free, but if we don't already have an `id` for something (ie, when we're creating it), we have to do it manually like this.

This is a very dense thing to try and communicate in this little comment box, so if I didn't make it clear enough, just let me know!

src/components/HappHosting/HappHosting.connector.js
```diff
@@ -24,20 +23,30 @@ const hostingUser = graphql(HostingUserQuery, {
...
     return {
-      // NOTE: Currently host_doc is not validated and the content is irrelevant to the dna...
-      registerHostingUser: ({host_doc}) => mutate({
+      // NOTE: Currently hostDoc is not validated and the content is irrelevant to the dna...
+      registerHostingUser: hostDoc => mutate({
         variables: {
-          host_doc
+          hostDoc
+        },
+        update: (cache, { data: { registerHostingUser } }) => {
```
robbiecarlton, line 34: this whole update function is the thing I was talking about "the last thing you haven't seen about Apollo". Basically, it's a callback that gets called after the mutation returns. it gets passed the result of the mutation, as well as the cache. The cache is the Apollo equivalent of the redux store, it's where all the local state is stored, specifically the results of all our graphQL queries.

So what I'm doing here is taking the result of the registerHostingUser mutation and writing it to the cache under the HostingUserQuery, so that when Apollo then goes to fetch the results of HostingUserQuery, it will return the result of registerHostingUser, and won't have to call the dna again to get the result.

Most often we get this kind of behavior for free, but if we don't already have an id for something (ie, when we're creating it), we have to do it manually like this.

This is a very dense thing to try and communicate in this little comment box, so if I didn't make it clear enough, just let me know!

src/components/HappHosting/HappHosting.js
```diff
@@ -5,17 +5,15 @@ import HashIcon from 'components/HashIcon'
...
-  console.log("Host WHOAMI Result >> ", hostingUser)
+  console.log('Host WHOAMI Result >> ', hostingUser)
 
   return <div>
-    {hostingUser && hostingUser.result ?
-       <h2>Current Host: {hostingUser.result}</h2>
-    :
-       <SpecificButton styleName='center-items' onClick={()=>registerHostingUser({host_doc:"info to register host"})} >Register as Host</SpecificButton>
-     }
+    {hostingUser && hostingUser.id
```
robbiecarlton, line 11: change from `.result` to `.id` because hosting user is now of type `HhaUser`, which has a single field `id`

src/graphql-server/dnaInterfaces/envoyInterface.js
```diff
@@ -0,0 +1,68 @@
+import axios from 'axios'
```
robbiecarlton, line 1: ignore this file for now, I haven't really touched it. Just moved the code into it's own file

src/graphql-server/dnaInterfaces/hhaDnaInterface.js
```diff
@@ -1,25 +1,29 @@
...
 export const INSTANCE_ID = 'hha' // holo-hosting-app
 const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
 
-export const HhaDnaInterface = {
+const HhaDnaInterface = {
   currentUser: {
-    create: (host_doc) => createZomeCall('host/register_as_host')({host_doc}),
-    get: () => createZomeCall('host/is_registered_as_host')(),
+    // create is getting passed a variable (hostDoc) but is currently ignoring it
+    create: async () => {
```
robbiecarlton, line 9: here and in `get` below, I'm transforming the data coming back from the zome so that it is the shape graphQL types expect. in this case, a javascript object with a single `id` field.

the `links` stuff is how the `host/is_registered_as_host` zome call returns its data

src/graphql-server/resolvers.js
```diff
@@ -11,7 +11,7 @@ export const resolvers = {
   Mutation: {
     registerUser: (_, userData) => dataMappedCall('person', userData, HyloDnaInterface.currentUser.create),
 
-    registerHostingUser: (host_doc) => HhaDnaInterface.currentUser.create(host_doc),
+    registerHostingUser: (_, hostDoc) => HhaDnaInterface.currentUser.create(hostDoc),
```
robbiecarlton, line 14: notice the change here. This will have been causing you problems. The first param to the resolver function is a `parent` object, which for these top level queries we typically (never?) need.

As well as writing resolvers for top level queries of Mutation and Query, you can also write resolvers for individual types. That's when the `parent` param becomes useful. But we'll cross that bridge when we come to it

src/graphql-server/schema.graphql
```diff
@@ -10,6 +10,10 @@ type HappStoreUser {
   name: String
 }
 
+type HhaUser {
```
robbiecarlton, line 13: as mentioned, defined the `HhaUser` type, and then used it below as the return type of `Query.hostingUser` and `Mutation.registerHostingUser`

I propose that we should as a rule (which I am currently breaking elsewhere but will fix) call the hc address of the entity`id` in the frontend, and then other hashes and addresses we can call `hash` or `address` as makes sense in the context.

src/mock-dnas/mockData.js
```diff
@@ -93,14 +93,13 @@ const data = {
   },
   hha: {
     host: {
-      register_as_host: args => {
-        console.log(" `host_doc` args passed when registering host; reqs for host registaration are currently being disregarded.) : ", args)
-        
-        data.hha.host.is_registered_as_host = { result: 'QmHHAHostRegistrationAddress' }
-        return data.hha.host.is_registered_as_host
+      register_as_host: () => {
```
robbiecarlton, line 96: this has been refactored so that it behaves like the dna (as far as I can tell. Haven't read the dna code, just fired it up and poked it a bit to see what it did).
let me know if you see anything I've missed or misunderstood here.

robbiecarlton (commented):
lmk if you have questions

PR #11, opened 2019-08-06
(Do not merge) hha integration code review step 2

robbiecarlton:


robbiecarlton:
This one is messier than step 1, but thought it would be helpful to continue doing these in chunks.

PR #13, opened 2019-08-09
As an HP Owner I can browse available hApps, and host and unhost them

robbiecarlton:


lorenjohnson (commented):
@robbiecarlton I've reviewed, and this looks just fine for now. Is there anything besides PR review and UI work holding this back from being merged.

src/components/RegisterUser/RegisterUser.connector.test.js
```diff
@@ -0,0 +1,44 @@
...
+    },
+    result: {
+      data: {
+        hostingUser: mockHostingUser
+      }
+    }
+  }
+]
+
+describe('connector', () => {
```
robbiecarlton, line 26: see comments in `RegisterUser.test.js` below

src/components/RegisterUser/RegisterUser.js
```diff
@@ -0,0 +1,22 @@
+import React, { useEffect } from 'react'
+import { get } from 'lodash/fp'
+
+export function RegisterUser (props) {
```
robbiecarlton, line 4: This component is mostly a placeholder. I expect our auth/login requirements to be more complicated than this. But for now, we have to make sure that the hostingUser is registered as a host or they can't see the available happs.

src/components/RegisterUser/RegisterUser.test.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
+import RegisterUser from './RegisterUser'
+import { render } from '@testing-library/react'
+
+const childrenText = 'The Children'
+const children = <div>{childrenText}</div>
+
+describe('RegisterUser', () => {
```
robbiecarlton, line 8: For this component it felt much better to test the component without the connector. This is an unusually complex component (three possible branches!), and I still currently think the integration style is better for simpler components. But this has reduced my confidence in that a little.

For now I think we just keep doing what seems right in a given case, and check back in regularly to see if there's a conclusion to come to.

lorenjohnson (commented):
LGTM 👍

PR #14, opened 2019-08-09
Basic CircleCi setup with Yarn

danekszy:


lorenjohnson (commented):
This looks great--one thing though: 

To avoid possible future issues we need to decide on whether we're using `yarn` or 'npm' for the dependency installation as well as the runner. In your config there is a mix of both. I really don't care which, nor am I remembering what we decided as a team in retreat on this. I think `yarn`? Whatever it is we should all be clear and doing the same thing in our local dev test runs.

danekszy:
> To avoid possible future issues we need to decide on whether we're using `yarn` or 'npm' for the dependency installation 

We're using npm here only for running scripts. Installation is handled by yarn (which TBH is suboptimal in this case). 😉

lorenjohnson (approved):
LGTM 🥇

PR #16, opened 2019-08-09
Merge pull request #15 from Holo-Host/hha-update-three

JettTech:
fully merged with working enable and disable buttons


robbiecarlton:
@JettTech can we coordinate asap. I'm working on a different branch so targetting hha-update-three is potentially causing you extra work.

PR #17, opened 2019-08-12
Add CI npm script, avoid double linting

danekszy:
Let's use `npm run ci` as a contract between `package.json` and CircleCI.
Linting task remains as was - `npm run test:lint`

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,10 @@
   "scripts": {
     "start": "rm -rf ./node_modules/.cache/babel-loader && node scripts/start.js",
     "build": "node scripts/build.js",
-    "test": "npm run test:lint && node scripts/test.js",
-    "test:lint": "standard"
+    "test": "npm run test:lint && npm run test:unit",
+    "test:lint": "standard",
+    "test:unit": "node scripts/test.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,8 +5,10 @@
   "scripts": {
     "start": "rm -rf ./node_modules/.cache/babel-loader && node scripts/start.js",
     "build": "node scripts/build.js",
-    "test": "npm run test:lint && node scripts/test.js",
-    "test:lint": "standard"
+    "test": "npm run test:lint && npm run test:unit",
+    "test:lint": "standard",
+    "test:unit": "node scripts/test.js",
+    "ci": "CI=1 npm run test:unit"
```
robbiecarlton, line 11: what does setting CI=1 do?

danekszy: Disables watch mode ;)
https://github.com/Holo-Host/hp-admin/blob/develop/scripts/test.js#L40

danekszy: BTW, looks to me there's no real reason to use test.js instead of simple npm script...

robbiecarlton (approved):
LGTM

PR #18, opened 2019-08-13
Add "Host" and "Unhost" to BrowseHapps

robbiecarlton:
Also added a little to the testing framework.

Specifically, the `mockData` is now reset after every test. This is so that each test knows exactly what to expect from `mockData`
This involves a `cloneDeep` of the whole tree after each test, which *may* start to be a noticeable performance issue as the tree gets bigger and we write more tests.

If we start to notice the tests taking a long time, we can remove the `afterEach` that's setup in the jest setup in `package.json` and only call `resetMockData` when the test is updating the data.

I'm losing a little of my enthusiasm for this testing pattern, and noticing leaning more towards traditional unit testing, but I'm not ready to give up yet :D


lorenjohnson (approved):
Looks good, go ahead and merge. 

My one niggle is that I might have held back on `react-toggle` until we are moving out of Ugly UI in a bit, but so be it. We'll need a control like that a few places and I'm assuming you vetted some options and felt this was better than the alternatives including rolling our own.

robbiecarlton:
@lorenjohnson oh, good catch. I'm actually not using it any more. It was for the Hosting Overview page which we took out. So I'm going to remove react-toggle.

PR #20, opened 2019-08-16
Remove mock data reset

robbiecarlton:
@JettTech @danekszy @lorenjohnson 
Tagging you in this just so you get a chance to see it, I'd be surprised if this change breaks anything you're doing, but if it does, please let me know.

I'm removing the code I added that resets the mock data after each test. It was causing a problem in using the mock data as originally intended: for manually testing the UI. The problem was soluble, but it started to feel increasingly like a bad smell. So for now, we will not be able to use the mock data to test mutations, which means leaning more on the unit testing pattern rather than the integration testing pattern.

LMK if you have thoughts or questions about this


src/graphql-server/resolvers.js
```diff
@@ -3,13 +3,27 @@ import HyloDnaInterface from './dnaInterfaces/hyloDnaInterface'
 import HappStoreDnaInterface, { getHappDetails } from './dnaInterfaces/happStoreDnaInterface'
```
robbiecarlton, line 3: the changes in this file were accidentally caught in this PR, you can ignore them they're not part of what I'm wanting you to look at.

PR #21, opened 2019-08-18
(do not merge. lame, autodated, legacy implementation. hocs. yawn) 5090 - Host Pricing

robbiecarlton:


src/graphql-server/dnaInterfaces/HhaDnaInterface.js
```diff
@@ -1,4 +1,7 @@
+import { isEmpty } from 'lodash/fp'
```
danekszy, line 1: I think we should settle on either `lodash/fp` or `lodash`. This leads to confusion and worse tree-shaking...

src/graphql-server/dnaInterfaces/HhaDnaInterface.js
```diff
@@ -1,4 +1,7 @@
...
 import { instanceCreateZomeCall } from '../holochainClient'
+import Promise from 'bluebird'
```
danekszy: Why bluebird over built-in promises?

src/graphql-server/dnaInterfaces/HhaDnaInterface.js
```diff
@@ -50,6 +54,34 @@ const HhaDnaInterface = {
         happStoreId,
         isEnabled: true
       })))
+  },
+
+  hostPricing: {
+    get: async () => {
+      // we need an id to call get_service_log_details, and because we set all apps the same in add_service_log_details, it doesn't matter which app the id comes from
+      const allAvailable = await HhaDnaInterface.happs.allAvailable()
+      if (isEmpty(allAvailable)) throw new Error("Can't set Host Pricing: no happs available to host.")
```
danekszy, line 62: nitpick: more whitespace would make for more readable code ;)

src/graphql-server/dnaInterfaces/HhaDnaInterface.js
```diff
@@ -50,6 +54,34 @@ const HhaDnaInterface = {
...
+      return createZomeCall('provider/get_service_log_details')({ app_hash: allAvailable[0].id })
```
danekszy, line 63: Zome address here looks like a perfect constant candidate to me

src/models/HostPricing.js
```diff
@@ -0,0 +1,6 @@
+export const UNITS = {
+  cpu: 'cpu',
```
danekszy, line 2: Uppercase values for better debugging readablity?

src/pages/BrowseHapps/BrowseHapps.js
```diff
@@ -7,8 +7,9 @@ import cx from 'classnames'
 export default function BrowseHapps ({ allAvailableHapps = [], enableHapp, disableHapp, history: { push } }) {
   const sortedHapps = allAvailableHapps.sort((a, b) => a.isEnabled ? -1 : b.isEnabled ? 1 : 0)
   const goToMenu = () => push('/menu')
+  const goToPricing = () => push('/pricing')
 
-  return <div>
+  return <div styleName='container'>
     <div styleName='header'>
       <span styleName='title'>hApps</span>
```
danekszy, line 14: Actual heading (`h1`, `h2` etc) for structure and accessibility?

src/pages/ManagePricing/ManagePricing.connector.js
```diff
@@ -0,0 +1,24 @@
+import { graphql } from '@apollo/react-hoc'
+import { flow as compose } from 'lodash'
+import HostPricingQuery from 'graphql/HostPricingQuery.gql'
+import UpdateHostPricingMutation from 'graphql/UpdateHostPricingMutation.gql'
+
+const hostPricing = graphql(HostPricingQuery, {
+  props: ({ data: { hostPricing } }) => ({ hostPricing })
```
danekszy, line 7: This will throw if any key here is missing. ` = {}` prevents that ;)

src/pages/ManagePricing/ManagePricing.connector.test.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+]
+
+describe('connector', () => {
+  it('runs the HostPricingQuery', async () => {
+    let hostPricing
+
+    const MockComponent = props => {
+      hostPricing = props.hostPricing
+      return null
```
danekszy, line 48: If explicit null isn't needed anywhere, you can just terminate the function

src/pages/ManagePricing/ManagePricing.js
```diff
@@ -0,0 +1,66 @@
...
+          <option value={value} key={value}>
+            {label}
+          </option>)}
+      </select>
+    </div>
+
+    <div styleName='price-input-wrapper'>
+      <label styleName='price-input-label'>
+        Holofuel per unit
+        <input type='text' value={pricePerUnit} onChange={onFuelInputChange} styleName='price-input' />
```
danekszy: What's the input format here? Maybe we could go with number or at least `pattern` for numeric keyboard on mobiles?

src/graphql-server/dnaInterfaces/HhaDnaInterface.js
robbiecarlton: agreed. My vote is `lodash/fp`, I've been trying to only use that.

robbiecarlton: did I use `lodash` somewhere?

src/graphql-server/dnaInterfaces/HhaDnaInterface.js
robbiecarlton: wanted `Promise.map`, but I can just use `Promise.all` so I'll remove it

src/pages/ManagePricing/ManagePricing.connector.test.js
robbiecarlton: actually you can't. You need to return null from a react component.

src/pages/ManagePricing/ManagePricing.connector.js
robbiecarlton: eh, we probably want it to throw in that case. Fail early.

src/pages/ManagePricing/ManagePricing.js
robbiecarlton: yeah, will update

src/graphql-server/dnaInterfaces/HhaDnaInterface.js
robbiecarlton: yeah, I've been thinking about that. Reason I haven't moved them into constants yet is because we mostly (/ever?) don't use them multiple times, so the dnaInterface files *are* the source of truth for the zome calls. I could go either way for now, and if and when we start needing them in multiple places, then they should go in constants.

A more ripe candidate is the routing paths. Just haven't got around to putting them in a file yet

src/graphql-server/dnaInterfaces/HhaDnaInterface.js
robbiecarlton: hmm. The Promise.all version is much more verbose and harder to read. Would prefer to keep bluebird for this. let's talk at standup

src/graphql-server/dnaInterfaces/HhaDnaInterface.js
danekszy: I saw it in the project previously ;)

src/pages/ManagePricing/ManagePricing.connector.test.js
danekszy: ah, right :)

PR #23, opened 2019-08-18
5090 - Apollo hooks version of ManagePricing feature

robbiecarlton:
attn: @lorenjohnson @danekszy @JettTech 

This is a demo of a different approach to doing "connected" (aka container) components.

Rather than having a separate connector file, we just use the apollo hooks `useQuery` and `useMutation` to make graphql calls.

Pros

- Easier to see where the values are coming from in the code
- Distinguishes Container from Dumb components in the code.
- Fewer files / less boiler plate
- Gives us access to mutation "loading" state without having to jump through weird hoops
- This is how Apollo wants us to do it. They are now calling all other approaches "legacy" and recommending using hooks.
- Specifically, this approach makes more sense than the HOC in a separate connector file approach wrt their recommended approach to testing Connected components.
- Removes the need for the very simplistic separate connector tests.

Cons

- Makes certain kinds of testing harder or impossible. Specifically, we can't test the container components separately from their connection to apollo. Apollo gives us MockedProvider so that this doesn't cause too many problems, but it does mean we can't just render a component with some jest.fn()s as props and test that they get called.
- Adds extra code to the render method, decreasing readability.

~~Note: right now I'm skipping the mutation test in the component because there is still some question of the best way to test apollo mutations. This isn't directly because of using hooks; in the non-hooks version of this feature I just don't have a test for the mutation in the connector. The difference is, I could still test the flow of the mutation getting _called_ in the component. Now I can't do that until we decide on the right way to test mutations.~~

Just added a mutation test.


danekszy:
We've already talked about this a bit... :) I think hooks can give us nicer reusability and a bit cleaner API - I disagree that it'll make render less readable. 😉

PR #24, opened 2019-08-19
Use npm & package-lock instead of yarn & yarn.lock

danekszy:
Should fail with yarn.lock present and should execute a bit faster (esp. linting stage)


robbiecarlton (approved):
LGTM. Thanks for throwing on yarn.lock, that will for sure save some headaches.

lorenjohnson (approved):
LGTM 👍

PR #27, opened 2019-08-19
Run test:unit instead of test in the test step of circleci

robbiecarlton:
The current setup means that bad linting causes *both* steps to fail. So the test step fails even if all the tests are fine.

Does this fix look right to you @danekszy?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ commands:
       - setup-nodejs
       - run:
           name: Test
-          command: npm run test:unit
+          command: npm run ci
   run-lint:
     steps:
       - prepare-env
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ commands:
       - setup-nodejs
       - run:
           name: Test
-          command: npm test
+          command: npm run test:unit
```
danekszy:
```suggestion
          command: npm run ci
```

danekszy:
Changed `test:unit` to `ci` like before. But yeah, somehow this bug made way into this version of CI flow.
Thanks for fixing that! 🙌

PR #29, opened 2019-08-21
5088 hApp details view

danekszy:


src/components/Button/Button.js
```diff
@@ -1,11 +1,29 @@
...
+  disabled = false,
+  wide = false,
+  variant = 'secondary',
+  className,
+  onClick = () => {},
+  children
+}) {
+  const classes = cx('button', {
+    wide,
+    primary: (variant === 'primary'),
```
robbiecarlton: This is cool, I've never seen it before.

robbiecarlton, line 15: at first I thought this was some fancy destructuring syntax and was like "woah".

I think I like this style

`<Button mini>blah</Button>`

more than

`<Button variant='mini'>blah</Button>`

Can see arguments for both.

src/components/HappThumbnail/HappThumbnail.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import './HappThumbnail.module.css'
+
+const HappThumbnail = ({ title, url, className }) => (
```
robbiecarlton, line 4: I would have held off on making this a component yet.

src/pages/BrowseHapps/BrowseHapps.js
```diff
@@ -34,15 +37,15 @@ export default function BrowseHapps ({ history: { push } }) {
           key={happ.id} />)}
     </div>}
 
-    <Button onClick={goToPricing} styleName='pricing-button'>Manage Pricing</Button>
+    <Button primary wide onClick={goToPricing} styleName='pricing-button'>Manage Pricing</Button>
```
robbiecarlton: The code as written is how I'd prefer we do it, but with Button as it currently is, shouldn't this be

`<Button variant='primary' wide`

src/pages/BrowseHapps/BrowseHapps.test.js
```diff
@@ -68,11 +70,24 @@ const mocks = [
...
+    history
+  }
+}
+
 describe('BrowseHapps', () => {
   it('renders', async () => {
     let getAllByRole
     await act(async () => {
-      ({ getAllByRole } = render(<MockedProvider mocks={mocks} addTypename={false}>
+      ({ getAllByRole } = renderWithRouter(<MockedProvider mocks={mocks} addTypename={false}>
```
robbiecarlton: why is this necessary?

src/pages/HappDetails/HappDetails.js
```diff
@@ -0,0 +1,84 @@
...
+import './HappDetails.module.css'
+
+import { useQuery, useMutation } from '@apollo/react-hooks'
+import AllAvailableHappsQuery from 'graphql/AllAvailableHappsQuery.gql'
+import EnableHappMutation from 'graphql/EnableHappMutation.gql'
+import DisableHappMutation from 'graphql/DisableHappMutation.gql'
+const findAppByID = id => happ => happ.id === id
+
+export default function BrowseHapps ({
+  history: { push } = {},
```
robbiecarlton: don't think we want a default here

src/pages/HappDetails/HappDetails.js
```diff
@@ -0,0 +1,84 @@
...
+  match: { params } = {}
+} = {}) {
+  const { data: { allAvailableHapps = [] } } = useQuery(AllAvailableHappsQuery)
```
robbiecarlton: could instead make a HappQuery that takes the id and just returns the Happ we want. In this case it def doesn't save loc, but I like it better for saying more exactly what we mean.

src/pages/HappDetails/HappDetails.test.js
```diff
@@ -0,0 +1,178 @@
...
+import { createMemoryHistory } from 'history'
+import { MockedProvider } from '@apollo/react-testing'
+import wait from 'waait'
+import HappDetails from './HappDetails'
+import { appOne as appHoloFuel } from 'mock-dnas/happStore'
+import AllAvailableHappsQuery from 'graphql/AllAvailableHappsQuery.gql'
+import EnableHappMutation from 'graphql/EnableHappMutation.gql'
+import DisableHappMutation from 'graphql/DisableHappMutation.gql'
+
+const enableHappMock = {
```
robbiecarlton: wondering about just importing all these mocks from `BrowseHapps.test.js` It's probably not right, just a thought.

src/pages/HappDetails/HappDetails.test.js
```diff
@@ -0,0 +1,178 @@
...
+) {
+  return {
+    ...render(<Router history={history}>{ui}</Router>),
+    history
+  }
+}
+
+async function renderHoloFuelApp (address = 'QmHHAHappEntryAddressHash1') {
+  const render = await renderWithRouter(<MockedProvider mocks={mocks} addTypename={false}>
+    <HappDetails history={{}} match={{ params: { address } }} />
```
robbiecarlton: repeating myself, but it's confusing to me to be using `renderWithRouter` here and yet also passing in an empty object for `history`.

robbiecarlton (commented):
Couple of questions for me, and I notice you are quicker to extract out components than I am, but overall it looks good.

src/components/Button/Button.js
danekszy: Your suggestion looks nicer, but doesn't prevent us from setting both `primary` and `secondary` options, which would break the design.
If there aren't any strong opinions, I'd stick to single prop (wide is a switch and not a variant).

src/components/HappThumbnail/HappThumbnail.js
danekszy: Yeah, this didn't help much, but it just looked obvious to me given the designs we have right now.
Now that it's done, I don't see a reason to revert, unless you feel strongly about it. ;)

src/pages/BrowseHapps/BrowseHapps.js
danekszy: Ha, nice catch! Thanks!

src/pages/BrowseHapps/BrowseHapps.test.js
danekszy: Mainly so that Link is not throwing errors. Sometimes it also handles actual navigation...

src/pages/HappDetails/HappDetails.test.js
danekszy: Ideally this would get extracted somewhere. Pulling one test from another seems like a bad idea. :P

src/pages/HappDetails/HappDetails.test.js
danekszy: Not sure if needed here. If not, I'll remove it for clarity.
Thanks!

src/pages/BrowseHapps/BrowseHapps.test.js
robbiecarlton: my preference would be to only bring it in when we know we need it, and to not pass in a dummy history object in those cases

src/components/HappThumbnail/HappThumbnail.js
robbiecarlton: don't feel strongly, I think it's fine to leave it like this. Just noting a slight difference in approach.

src/pages/BrowseHapps/BrowseHapps.test.js
danekszy: I agree, though sometimes you need dummy history anyways - Router itself doesn't inject the prop.

src/components/Button/Button.js
robbiecarlton: This makes sense.

hmm. so 'wide' is a switch, but 'mini' is a variant? This seems like a gotcha.

src/pages/BrowseHapps/BrowseHapps.test.js
robbiecarlton: ah, that's surprising to me

src/components/Button/Button.js
danekszy: Yeah. I mean mini might be a wrong name for a variant here, but I know the `wide` modifier is quite common with buttons, and making separate variants for it would be cumbersome.

src/pages/BrowseHapps/BrowseHapps.test.js
danekszy: it's because renderWithRouter supplies some generic in-memory router (so there is context for `Link`s to work with) but we're not using our whole router here.
Usually it's router that renders the component, while here (in tests) we're providing the component wrapped in mocked providers and what not. 😉

src/pages/HappDetails/HappDetails.test.js
danekszy: I'd leave it as it is right now. If there are more cases like this, perhaps we can just setup some space in project for test related utils etc.

src/pages/HappDetails/HappDetails.test.js
danekszy: Removed replaced with plain render wherever I can, but history object needs to stay here (unless we want to put this as a prop default)

src/graphql-server/schema.graphql
```diff
@@ -37,6 +37,7 @@ type Query {
   happStoreUser: HappStoreUser
   hostingUser: HhaUser
   happs: [Happ]
+  happByStoreId(storeId: String!): Happ
```
lorenjohnson: For our API I'd like this to be simply `happ` or optionally--and perhaps best for now--extend `happs` to receive an and query on an optional `id` (or `happStoreId` if that is really the best way). If you go the `happ` way then just implement the resolver like the non-array version of `happs`.

Either is fine with me, and I'm happy to pair with you on how to achieve this.

src/pages/HappDetails/HappDetails.js
```diff
@@ -0,0 +1,98 @@
...
+export default function BrowseHapps ({
+  history: { push },
+  match: { params }
+} = {}) {
+  const q = useQuery(HappByStoreIdQuery, {
+    variables: {
+      storeId: params.happStoreId || ''
+    }
+  })
+  console.log('hbi', params.happStoreId,  q)
```
lorenjohnson: `console.log` ... but I think this is here because this is still WIP. If so please reflect that in the title (only important once you tag for review :) )

src/pages/PrimaryLayout/PrimaryLayout.js
```diff
@@ -15,9 +16,9 @@ export function PrimaryLayout () {
 
     <Route path='/(|dashboard)' exact component={Dashboard} />
     <Route path='/menu' component={MainMenu} />
-    <Route path='/browse-happs' component={BrowseHapps} />
+    <Route path='/browse-happs' exact component={BrowseHapps} />
+    <Route path='/browse-happs/:happStoreId' component={HappDetails} />
```
lorenjohnson: Once you update the resolvers please make this parameter simply `happId` or `id` and reflect this through.

lorenjohnson (changes requested):
Tests and linting are currently failing, also see notes. Otherwise LGTM 👍

src/pages/PrimaryLayout/PrimaryLayout.js
danekszy: Renamed to `appId` actually. Would this suffice for you?

src/graphql-server/schema.graphql
danekszy: Currently its `happById`. Will change then...

src/pages/PrimaryLayout/PrimaryLayout.js
lorenjohnson: Yep, that's fine 👍 

As a side note: We should all sync about the use of `happ` vs `app` in different contexts at some point. There doesn't seem to be consistency.

danekszy: Agreed

src/pages/HappDetails/HappDetails.test.js
```diff
@@ -0,0 +1,110 @@
...
+import mockEnvoyInterface from 'data-interfaces/EnvoyInterface'
+import hhaInterface from 'data-interfaces/HhaDnaInterface'
+
+jest.mock('data-interfaces/EnvoyInterface')
+
+afterEach(() => {
+  apolloClient.resetStore()
+})
+
+async function renderHoloFuelApp (appId = 'QmHHAHappEntryAddressHash1') {
```
lorenjohnson: Shouldn't this be `renderHappDetails` instead of `renderHoloFuelApp` throughout?

src/pages/HappDetails/HappDetails.test.js
```diff
@@ -0,0 +1,110 @@
+import React from 'react'
+import { render, fireEvent, act, cleanup } from '@testing-library/react'
+import { ApolloProvider } from '@apollo/react-hooks'
+import apolloClient from 'apolloClient'
+import wait from 'waait'
+import HappDetails from './HappDetails'
```
lorenjohnson: Super small thing, but I suggest we stick with leaving the most specific thing as the last import. So `HappDetails` would go to the last import entry.

lorenjohnson (approved):
I'm ambivalent about the confirmation modal as it was purposely left out of the spec on this for this release. You can leave it in, I don't want you feel the work was wasted, but honestly I would slightly prefer it not in so that the current behaviour was the same between the hApp index and this screen (no confirmation).

Also, for these early lofi releases like this I suggest we resist abstracting shared components (e.g. `HappThumbnail`, `HostButton`), but if you then I would want us to go ahead and implement put them in the other places they would currently be used (hApp index screen in this case). Does that make sense? Again, if honest I'd prefer they just be brought back into the detail component file for now and evaluated for abstraction on the hifi or UI design round. 

Approving as I'd like to keep us moving, but if you are able to address these things and or want to discuss that'd be ideal.

PR #35, opened 2019-09-03
8378 - As an HP Owner I can login with my provided email and password when I return to HoloPort Admin

robbiecarlton:
One possible addition would be to have the app redirect to this page for the demo. But I figured that was not so important given that the whole mechanism will be different once we actually implement auth.

Also, per the designs this has the same header (burger menu, title and avatar button) as every other page, but that seems likely to change in later versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6071,6 +6071,11 @@
         "readable-stream": "^2.3.6"
       }
     },
+    "fn-name": {
+      "version": "2.0.1",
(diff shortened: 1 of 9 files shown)


lorenjohnson (approved):
LGTM 👍  I went ahead and added field labels as that was what I expected and want to setup things to consider form styling across the board for now. Merge away...

robbiecarlton:
@lorenjohnson Ah, there were no labels in the design so I didn't put any in there.

PR #37, opened 2019-09-03
UI work related to: 4705 HoloFuel Plumbing (for reference -- don't merge)

JettTech:


package.json
```diff
@@ -109,6 +111,7 @@
     "webpack": "4.29.6",
     "webpack-dev-server": "3.2.1",
     "webpack-manifest-plugin": "2.0.4",
+    "webpack-node-externals": "^1.7.2",
```
robbiecarlton: let's uninstall this until we know we're going to use it.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -0,0 +1,197 @@
...
+    allActionable: async () => {
+      const { requests, promises } = await createZomeCall('transactions/list_pending')()
+      const actionableTransactions = requests.map(presentPendingRequest).concat(promises.map(presentPendingOffer))
+      return actionableTransactions.sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
+    },
+    allWaiting: async () => {
+      const { transactions } = await createZomeCall('transactions/list_transactions')()
+      const listOfNonActionableTransactions = transactions.map(presentTransaction)
+      // NOTE: Filtering out duplicate IDs should prevent an already completed tranaction from displaying as a pending tranaction if any lag occurs in data update layer.
+      const noDuplicateIds = [...new Set(listOfNonActionableTransactions)]
```
robbiecarlton: does `new Set` specifically use `id` to filter duplicates? I'm not familiar with it. 
Any consideration to using `lodash/uniqBy`? with `uniqBy` we can be explicit about our duplication criteria

src/data-interfaces/__mocks__/HoloFuelInterface.js
```diff
@@ -0,0 +1,16 @@
...
+  transactions: {
+    allComplete: jest.fn(() => ['mockCompleteTransactionOne', 'mockCompleteTransactionTwo']),
+    allActionable: jest.fn(() => ['mockActionableTransactionOne', 'mockActionableTransactionTwo']),
+    allWaiting: jest.fn(() => ['mockWaitingTransactionOne', 'mockWaitingTransactionTwo'])
+  },
+  requests: {
+    create: jest.fn()
+  },
+  offers: {
+    create: jest.fn(),
```
robbiecarlton: Just an FYI, as it's not being tested yet, but ultimately these mocks should be returning a transaction object that reflects their input data. This will allow us to test the UI when eg creating a new request, that it updates the in the right place.

src/graphql-server/resolvers.js
```diff
@@ -19,13 +20,21 @@ export const resolvers = {
...
 
+    hostPricing: () => HhaDnaInterface.hostPricing.get(),
+
+    holofuelWaitingTransactions: () => HoloFuelDnaInterface.transactions.allWaiting(),
+
+    holofuelActionableTransactions: () => HoloFuelDnaInterface.transactions.allActionable(),
+
+    holofuelCompleteTransactions: () => HoloFuelDnaInterface.transactions.allComplete(),
+
+    holofuelLedgerState: () => HoloFuelDnaInterface.ledger.all(),
```
robbiecarlton: can we just call this `holofuelLedger`?

src/graphql-server/resolvers.js
```diff
@@ -54,7 +63,25 @@ export const resolvers = {
       return getHappDetails(happ)
     },
 
-    updateHostPricing: (_, { units, pricePerUnit }) => HhaDnaInterface.hostPricing.update(units, pricePerUnit)
+    updateHostPricing: (_, { units, pricePerUnit }) => HhaDnaInterface.hostPricing.update(units, pricePerUnit),
+
+    holofuelRequest: async (_, { counterparty, amount }) => {
+      const newRequest = await HoloFuelDnaInterface.requests.create(counterparty, amount)
+      // Logging this out for demo:
+      console.log('The new HF Request transaction body : ', newRequest)
```
robbiecarlton: do we still need these logs?

src/graphql-server/resolvers.test.js
```diff
@@ -32,6 +34,14 @@ describe('resolvers', () => {
         expect(mockGetHappDetails.mock.calls.map(c => c[0])).toEqual(['mockHappOne', 'mockHappTwo'])
       })
     })
+
+    describe('.holofuel', () => {
+      it('calls HoloFuelInterface.transactions.allComplete', async () => {
```
robbiecarlton: would be good to get tests for all the endpoints we added here.

I think given the relative complexity of the code it would also be good to get good coverage of the `HoloFuelDnaInterface`. I'm not sure exactly the best way to do that. Maybe unit tests with mocking of `createZomeCall`?  LMK if you want touch base on that to think it through together.

src/graphql-server/schema.graphql
```diff
@@ -32,13 +32,51 @@ type HostPricing {
...
+}
+
+type Transaction {
+  id: ID # ID will be the originCommitHash
+  amount: Float
+  counterparty: String
+  direction: TransactionDirection
+  status: TransactionStatus
+  type: TransactionType
+  timestamp: String # QUESTION: Can we use GraphQLDateTime here instead? This timestamp is actually received as an ISO8601...
```
robbiecarlton: It's probably right to use the most specific type available. I tend to get lazy and just use String for everything but let's try the more correct way.

src/graphql-server/schema.graphql
```diff
@@ -47,4 +85,8 @@ type Mutation {
   enableHapp(appId: String): Happ
   disableHapp(appId: String): Happ
   updateHostPricing(units: String, pricePerUnit: String): HostPricing
+  holofuelRequest(counterparty:String, amount:Float): Transaction
+  holofuelOffer(counterparty:String, amount:Float, requestId:ID): Transaction
+  holofuelAcceptOffer(transactionId: ID): Transaction # NOTE: holofuelAcceptOffer DOES NOT return the amount or counterparty. IF need these values, query from cache, as they will not change via this API.
+  holofuelRejectOffer(transactionId: ID): Transaction # NOTE: holofuelAcceptOffer DOES NOT return the amount or counterparty. IF need these values, query from cache, as they will not change via this API.
```
robbiecarlton: is this comment true of `holofuelRejectOffer`? if so can you update the comment to fix the mutation name. How do we handle rejecting requests? Is there a zome function for that? Or if this same mutation handles both, can we call it `holofuelRejectTransaction`

src/mock-dnas/holofuel.js
```diff
@@ -0,0 +1,302 @@
...
+  ]
+}
+
+const NO_SALT_ROUNDS = 10
+const holofuel = {
+  transactions: {
+    ledger_state: () => transactionList.ledger,
+    list_transactions: () => transactionList,
+    list_pending: () => pendingList,
+    request: ({ from, amount, deadline }) => bcrypt.hashSync((from + amount + deadline), NO_SALT_ROUNDS), // 'Qm1MNMQcEsd3BkQpaFUyZrViQ26axooErWtc', // NOTE: import a encryption to hash these for deterministic testing
```
robbiecarlton: we can remove all these comments

src/utils/index.js
```diff
@@ -10,3 +13,16 @@ export async function promiseMap (array, fn) {
   const resolved = await Promise.all(promiseArray)
   return resolved
 }
+
+export function formatDateTime (isoDate) {
```
robbiecarlton: I'm not sure this belongs in utils as its potentially pretty specific to the component you're using it in. Let's keep it in the component until we find that we need it somewhere else.

src/pages/HoloFuel/index.js
```diff
@@ -0,0 +1,8 @@
+// import Component from './HoloFuelTxOverview';
```
robbiecarlton: The directory structure here needs a tweak. Each page should have its own directory `PageComponentName` with children `index.js`, `PageComponentName.js`, `PageComponentName.module.css` etc. I see the value in nesting them all under a master `HoloFuel` directory to keep them separate, but as we haven't figured out how we're organizing the HoloFuel specific code I think for now just having those directories in the top level is fine.

src/pages/HoloFuel/HoloFuelDashboard.test.js
```diff
@@ -0,0 +1,31 @@
...
+      ({ getByText, getAllByRole } = render(<ApolloProvider client={apolloClient}>
+        <HoloFuelDashboard {...props} />
+      </ApolloProvider>))
+      await wait(0)
+    })
+
+    const listItems = getAllByRole('heading')
+    expect(listItems).toHaveLength(7)
+
+    // expect(getByText(transactionList.ledger.balance)).toBeInTheDocument()
```
robbiecarlton: remove comments

robbiecarlton (changes requested):
Overall looks great, good work, this is a gnarly feature. 

A few changes to look at. Probably the most involved thing left is testing.

Also there will likely be some additional work to be done when creating and updating transactions to maintain the cache (EG, when I create a request, it should show up in my list of 'waiting' transactions without having to refetch the list). I think we can look at that in the next pass.

src/graphql-server/schema.graphql
JettTech: There is an reject endpoint that specifically rejects an offered payment, not a request.  It currently takes in the same parameters as the receive_payment endpoint such that, like receive_payment, the reject endpoint in the holofuel DNA will need to be updated to only take the ID instead.

For reference, here is the current endpoint : 
```
    reject: {
            inputs: |promise: Promise, promise_sig: Signature, promise_commit: Address,
                     reason: Option<String>|,
            outputs: |result: ZomeApiResult<Address>|,
            handler: transaction::reject_handler
        }
```

I will update the comment.

JettTech: The option to cancel a request has been discussed; it just needs to be implemented.

src/pages/HoloFuel/index.js
JettTech: Thanks for this clarification. :rocket:  This is one of the questions I had and wished to discuss together.  I was thinking either all at top level, as you suggested, or within nested subdirs.  Anyway, I'll move them to the top-level for now.

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: I like lodash/uniqBy and have replaced it in the code. :raised_hands:  :100:  FYI: New Set basically allows unique values, all duplicates will be removed from the array provided.

src/graphql-server/resolvers.test.js
JettTech: Totally agree, and they are underway.  I'l touch base with you about them soon.

src/graphql-server/schema.graphql
JettTech: Let's hold on to this one to refer to soon. In order to reference that type, we need to import a specific module, which is causing some errors.

lorenjohnson (changes requested):
@JettTech we need to remove all of the UI related stuff from this branch and make sure the tests are passing and that the test coverage on the HoloFuel DNA and resolvers is complete. 

Please note that the HoloFuel DNA plumbing work itself is of priority to get ready and merged as we are getting into the actual app this sprint. 

Ping Robbie or I and one of us can help divide-up the branches appropriately and highlight any remaining work to getting this merge. 

---

We won't be directly using any of this UI code you've written here in the work we're currently doing, but I would encourage you to create a branch in which it is retained so either @robbiecarlton or I can do a separate PR review on it and with you as an education piece.  * It may also be a useful reference for us all as we build the into the HoloFuel app feature stories this sprint.

robbiecarlton:
@JettTech - is this PR ready to be closed now?

PR #19, opened 2019-08-16
3321 device settings

lorenjohnson:

PR #38, opened 2019-09-04
5261 my profile

danekszy:
Didn't add the modal with TOS.

Some more tests coming up

### DIFF
--- a/src/components/HashAvatar/HashAvatar.js
+++ b/src/components/HashAvatar/HashAvatar.js
@@ -1,0 +1,16 @@
+import React from 'react'
+import { Network as Identicon } from 'react-identicon-variety-pack'
+
+import './HashAvatar.module.css'
+
+const HashAvatar = ({ avatarUrl, email, size = 96, className }) => avatarUrl
+  ? <img
+    src={avatarUrl}
+    alt='Avatar'
+    styleName='image'
+    className={className}
+    style={{ width: size, height: size }}
+  />
+  : <Identicon seed={email || 'default-seed'} className={className} size={size} circle />
+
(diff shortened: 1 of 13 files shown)


src/components/HashAvatar/HashAvatar.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { Network as Identicon } from 'react-identicon-variety-pack'
+
+import './HashAvatar.module.css'
+
+const HashAvatar = ({ avatarUrl, email, size = 96, className }) => avatarUrl
```
robbiecarlton, line 6: ultimately I think we should be using agent hash for the avatar seed.

src/pages/MyProfile/MyProfile.js
```diff
@@ -0,0 +1,89 @@
...
+          You need to set your name.
+        </small>}
+      </label>
+
+      <label styleName='field'>
+        <span styleName='field-name'>Email</span>
+        <input
+          name='email'
+          placeholder='eg. alice@example.com'
+          ref={register({ required: true, pattern: EMAIL_REGEXP })}
```
robbiecarlton, line 58: If you set the input type to `email`, I think react-form-hooks automatically validates it. 
If not, we're using the Yup library for validation in device settings https://github.com/Holo-Host/hp-admin/pull/19 
Let's use one of those options rather than maintain our own email regex.

robbiecarlton (changes requested):
Only thing I think needs changing is the validation in the profile page, but it's not blocking so Loren is going to merge this for the demo.

PR #42, opened 2019-09-05
HoloFuel DNA plumbing

lorenjohnson:


src/data-interfaces/__mocks__/HoloFuelInterface.js
```diff
@@ -0,0 +1,31 @@
+import { makeIsoStringDateTime } from 'utils'
+
+const successfulTransactionResponse = ({ transactionId, amount, counterparty, status, type }) => {
```
robbiecarlton, line 3: 👍

src/graphql-server/resolvers.js
```diff
@@ -19,13 +20,21 @@ export const resolvers = {
...
 
+    hostPricing: () => HhaDnaInterface.hostPricing.get(),
+
+    holofuelWaitingTransactions: () => HoloFuelDnaInterface.transactions.allWaiting(),
+
+    holofuelActionableTransactions: () => HoloFuelDnaInterface.transactions.allActionable(),
+
+    holofuelCompleteTransactions: () => HoloFuelDnaInterface.transactions.allComplete(),
+
+    holofuelLedger: () => HoloFuelDnaInterface.ledger.all(),
```
robbiecarlton: this endpoint should probably be called `get` rather than `all`. The pattern is `all` endpoints return an array of objects of the same type, while `get` endpoints return a single object.

robbiecarlton (changes requested):
One small change, once that's done it's good to merge.

src/graphql-server/resolvers.js
JettTech: Good point.  I was thinking of it as 'all ledger details' rather than just 'balance' or 'fees,' and was therefore going back and forth between the two, but the data struct returned is still an obj, so I like having that be the deciding factor. :+1:

PR #43, opened 2019-09-06
6710 - An independent HoloFuel MVP hApp UI can be developed within the HoloPort Admin codebase

robbiecarlton:


src/pages/holofuel/PrimaryLayout/PrimaryLayout.js
```diff
@@ -0,0 +1,18 @@
+import React, { useContext } from 'react'
```
robbiecarlton, line 1: considered calling these components `HFPrimaryLayout`, `HFDashboard`, `HFHeader` etc, so that it's easier to distinguish between them when doing fuzzy search, but in the end I decided not to. I think it will just cause more mess than not, especially when separating this out into its own repo.

src/index.js
```diff
@@ -6,13 +6,33 @@ import { useMediaPredicate } from 'react-media-hook'
...
 import RegisterUser from 'components/RegisterUser'
 import ScreenWidthContext from 'contexts/screenWidth'
-import './global-styles/normalize.css'
-import './global-styles/colors.css'
-import './index.css'
+
+const HOLOFUEL_APP = true
 
 export function App () {
+  if (HOLOFUEL_APP) {
```
robbiecarlton, line 16: using a dumb constant rather than routing or query string or something because it's very simple and does the job :)

Using anything else would potentially increase the  complexity of extracting HoloFuel into its own repo.

src/pages/holofuel/PrimaryLayout/PrimaryLayout.js
lorenjohnson: agree, good call.

src/index.js
lorenjohnson: Sound good. This serves us this sprint as we'll be focused I think entirely on getting the HoloFuel MVP going. We'll revisit later.

lorenjohnson (approved):
LGTM 👍

PR #46, opened 2019-09-10
4705 holofuel data layer updated

JettTech:
Updates the HF data layer to include the following updates : 
 - whoami endpoint updates - now accepts an opt. agentId (aka. agent pubkey) param that will return the matching nick
- accept_request gql query - now calls the updated dna endpoint 'accept_requests_pending', which officially accepts the origin hash as the id
- resulting_balance key within adjustment obj is mocked for each transaction returned from list_transactions endpoint.  This is currently a WIP in the DNA.

### DIFF
--- a/src/data-interfaces/HoloFuelDnaInterface.js
+++ b/src/data-interfaces/HoloFuelDnaInterface.js
@@ -1,4 +1,5 @@
 import _ from 'lodash'
+import { mapValues} from 'lodash/fp'
 import { instanceCreateZomeCall } from '../holochainClient'
 
 export const currentDataTimeIso = () => new Date().toISOString()
@@ -90,12 +91,8 @@ function presentTransaction (transaction) {
   const stateStage = state.split('/')[1]
   const stateDirection = state.split('/')[0] // NOTE: This returns either 'incoming' or 'outgoing,' wherein, 'incoming' indicates the recipient of funds, 'outgoing' indicates the spender of funds.
   // NOTE: *Holofuel does NOT yet provide a balance that represents the 'RESULTING ACCT BALANCE after this transaction adjustment', instead of the only the tx adjustment balance or real-time balance.*
-  const adjustmentValues = Object.values(adjustment).map(key => key.Ok)
-  const adjustmentKeys = Object.keys(adjustment)
-  const parsedAdjustment = adjustmentValues.reduce((result, field, index) => {
-    result[adjustmentKeys[index]] = field
-    return result
-  }, {})
+  const parsedAdjustment = mapValues('Ok', adjustment)
+
   switch (stateStage) {
     case 'completed': {
       if (event.Receipt) return presentReceipt({ origin, event, stateDirection, eventTimestamp: timestamp.event, fees: parsedAdjustment.fees, presentBalance: parsedAdjustment.resulting_balance })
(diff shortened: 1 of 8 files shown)


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -173,7 +192,7 @@ const HoloFuelDnaInterface = {
 
     // NOTE: Below we reflect our current change to the receive_payment API; the only param should now be the transaction's origin id
     accept: async (transactionId) => {
-      await createZomeCall('transactions/receive_payment')({ origin: transactionId })
+      await createZomeCall('transactions/receive_payments_pending')({ origin: transactionId })
```
robbiecarlton, line 192: Doesn't the param need to be an array?

Looking at the HF PR it looks like the call should look like `({ promises: [transactionId] })`.

src/data-interfaces/__mocks__/HoloFuelInterface.js
```diff
@@ -1,14 +1,18 @@
...
     id: transactionId,
     amount: amount || 0,
     counterparty: counterparty || 'Data not avail until next refetch, retrieve from gql cache...',
     direction: type === 'offer' ? 'outgoing' : 'incoming',
     status,
     type,
-    timestamp: currentDataTimeIso
+    timestamp: currentDataTimeIso,
+    // NOTE: the following details are ONLY available in the 'completed transactions'...
+    fees: fees || 0,
```
robbiecarlton, line 13: we typically don't want to provide defaults for these as they will overwrite the cache. Can explain more on our call today.

robbiecarlton (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -87,10 +89,17 @@ function presentTransaction (transaction) {
   const { state, origin, event, timestamp, adjustment } = transaction
   const stateStage = state.split('/')[1]
   const stateDirection = state.split('/')[0] // NOTE: This returns either 'incoming' or 'outgoing,' wherein, 'incoming' indicates the recipient of funds, 'outgoing' indicates the spender of funds.
+  // NOTE: *Holofuel does NOT yet provide a balance that represents the 'RESULTING ACCT BALANCE after this transaction adjustment', instead of the only the tx adjustment balance or real-time balance.*
+  const adjustmentValues = Object.values(adjustment).map(key => key.Ok)
+  const adjustmentKeys = Object.keys(adjustment)
+  const parsedAdjustment = adjustmentValues.reduce((result, field, index) => {
```
robbiecarlton: you can use `lodash.mapValues here`

something like 
`import { mapValues} from 'lodash/fp'` <-- note the fp

`const parsedAdjustment = mapValues('Ok', object)`

or if that doesn't work

`const parsedAdjustment = mapValues(key => key.Ok, object)`

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Good thought. It actually is an param enum that can accepts either an array or a single hashString.
```
pub enum ListOrigins {
    Single(Address),
    Multiple(Vec<Address>),
}
```

src/data-interfaces/__mocks__/HoloFuelInterface.js
JettTech: Just for completion tot his comment, we discussed that this is generally an anti-pattern for Apollo in the DNA Interfaces, but not a problem in the DNA Interface **mocks**, as it doesn't affect the store / cache. I will go ahead and leave this one be.

PR #47, opened 2019-09-10
0753 - As a HoloFuel User I can view and take actions on an Inbox of incoming offers and requests

robbiecarlton:


src/pages/holofuel/Inbox/Inbox.test.js
```diff
@@ -0,0 +1,179 @@
...
+    let getAllByRole
+
+    await act(async () => {
+      ({ getAllByRole } = render(<ApolloProvider client={apolloClient}>
+        <Inbox />
+      </ApolloProvider>))
+      await wait(15)
+    })
+
+    const listItems = getAllByRole('listitem')
```
robbiecarlton, line 40: @JettTech - hey, this is potentially a little bit more elegant way of doing some of what we were doing in those other tests. Rather than using a `data-testid`, we use an aria role, which means we get a bit more accessibility and the same pattern for testing.

src/pages/holofuel/Inbox/Inbox.test.js
```diff
@@ -0,0 +1,179 @@
...
+    expect(listItems).toHaveLength(2)
+
+    listItems.forEach((item, index) => {
+      const { getByText } = within(item)
+      expect(getByText(actionableTransactions[index].notes)).toBeInTheDocument()
```
robbiecarlton: and I think this is an acceptable way to test if something is in your row, without having to do the extra layer of `within()` that we used in the tests for your components.

No need to update, just wanted to keep you posted on what I'm thinking re RTL.

robbiecarlton, line 45: @JettTech and I think this is an acceptable way to test if something is in your row, without having to do the extra layer of `within()` that we used in the tests for your components.

No need to update, just wanted to keep you posted on what I'm thinking re RTL.

lorenjohnson (approved):
As long as you guys are settled on the conclusions in the conversation in this PR, LGTM 👍

PR #50, opened 2019-09-12
7950 - HoloFuel Data Layer backend refactoring for how we re-pull transaction data

robbiecarlton:
This is currently pointed at `0753-inbox` branch for readability. Wait for that branch to be reviewed and merged before merging this.

### DIFF
--- a/src/data-interfaces/HoloFuelDnaInterface.js
+++ b/src/data-interfaces/HoloFuelDnaInterface.js
@@ -143,6 +143,15 @@ const HoloFuelDnaInterface = {
     }
   },
   transactions: {
+    getPending: async transactionId => {
+      const { requests, promises } = await createZomeCall('transactions/list_pending')({ origins: transactionId })
+      const transactionArray = requests.map(presentPendingRequest).concat(promises.map(presentPendingOffer))
+      if (transactionArray.length === 0) {
+        throw new Error(`no pending transaction with id ${transactionId} found.`)
+      } else {
+        return transactionArray[0]
+      }
+    },
     allComplete: async () => {
       const { transactions } = await createZomeCall('transactions/list_transactions')()
       const listOfNonActionableTransactions = transactions.map(presentTransaction)
@@ -162,8 +171,10 @@ const HoloFuelDnaInterface = {
       return noDuplicateIds.filter(tx => tx.status === 'pending').sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
     },
     decline: async transactionId => {
+      const transaction = await HoloFuelDnaInterface.transactions.getPending(transactionId)
       await createZomeCall('transactions/decline')({ origin: transactionId })
       return {
+        ...transaction,
         id: transactionId,
         status: STATUS.rejected,
         direction: DIRECTION.incoming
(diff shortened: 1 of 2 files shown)


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -143,6 +143,15 @@ const HoloFuelDnaInterface = {
     }
   },
   transactions: {
+    getPending: async transactionId => {
```
JettTech, line 146: Do we want this endpoint to *only* serve as a fetch for a single pending transaction?  If so, could we rename this function to reflect that specificity? (ie: 'getSinglePending')

JettTech (approved):
Overall, this is a solid implementation of what we discussed; looks great.  :100: 

I did make a single request to revisit and consider renaming the 'getPending' endpoint, but I don't perceive this as a blocker.  Let me know if you'd like to discuss this together further.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -162,8 +171,10 @@ const HoloFuelDnaInterface = {
       return noDuplicateIds.filter(tx => tx.status === 'pending').sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
     },
     decline: async transactionId => {
+      const transaction = await HoloFuelDnaInterface.transactions.getPending(transactionId)
       await createZomeCall('transactions/decline')({ origin: transactionId })
       return {
+        ...transaction,
         id: transactionId,
         status: STATUS.rejected,
         direction: DIRECTION.incoming
```
JettTech: The direction could actually be incoming or outgoing, as it is possible to decline both a request and an offer.  Let's just maintain the direction value returned in the 'transaction' object.

NB: Direction in this context indicates the final directional displacement of the funds, not the direction of a transaction stage.  (ie. A spender could be the recipient of a request, but not the recipient of the funds.)

JettTech (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -200,27 +211,16 @@ const HoloFuelDnaInterface = {
 
     // NOTE: Below we reflect our current change to the receive_payment API; the only param should now be the transaction's origin id
     accept: async (transactionId) => {
+      const transaction = await HoloFuelDnaInterface.transactions.getPending(transactionId)
       await createZomeCall('transactions/receive_payments_pending')({ origin: transactionId })
       return {
+        ...transaction,
```
JettTech, line 216: Similar to the 'decline' endpoint, let's just take the 'type', 'direction', and 'timestamp' from the values returned in 'transaction' object.

src/data-interfaces/HoloFuelDnaInterface.js
robbiecarlton: the convention we've been using is `get` endpoints return a single object, and `all` endpoints return a list.

src/data-interfaces/HoloFuelDnaInterface.js
robbiecarlton: ah, good catch, thanks.

src/data-interfaces/HoloFuelDnaInterface.js
robbiecarlton: 👍

robbiecarlton: actually, I think I should leave direction and type in here (but switch to using the timestamp from the fetched transaction as you suggest).

As I understand it, this endpoint specifically accepts an incoming offer, so we always know the type and direction. Having it be hardcoded here rather than implicit in the fetched transaction makes the code more self-documenting.

JettTech: Correct, for this endpoint, we will always now the type and direction.  I do actually like the idea of hard-coding this in for the purposes of readability and documentation. :100: Let's go with this approach.

JettTech (approved):

PR #53, opened 2019-09-17
7048 - layout architecture reorg

danekszy:
- [x] Separate layout from routing
- [x] Unified Header for HPA
- [x] Unified Header for HF
- [ ] Visuals (page paddings etc.)
- [ ] Connecting Header to Apollo - take Identicon seed from identity
- [ ] Unit-test connected Header


src/components/layout/PrimaryLayout.js
```diff
@@ -0,0 +1,22 @@
...
+import styles from './PrimaryLayout.module.css' // eslint-disable-line no-unused-vars
+import 'global-styles/colors.css'
+import 'global-styles/index.css'
+
+export function PrimaryLayout ({ children, footer }) {
+  const isWide = useContext(ScreenWidthContext)
+
+  return <div styleName={cx('styles.primary-layout', { 'styles.wide': isWide }, { 'styles.narrow': !isWide })}>
+    {children}
+    {footer && footer()}
```
danekszy: We're using render prop pattern here, to invert dependency tree here.

- Footer is what I'd call a "slot", and in this case is optional (in the end it will position the content at the bottom, like existing buttons).
  - Slot might take some arguments (here there's none) - eg. we might want to supply a class name that content provided by page would use.
- "props.children" might be considered the very minimal example and required slot.

src/pages/MyProfile/MyProfile.js
```diff
@@ -26,7 +28,18 @@ const MyProfile = ({
     setTosOpen(true)
   }
 
-  return <>
+  return <PrimaryLayout
```
danekszy, line 30: Here layout component actually helps us, so that we don't have to use fragments :)

src/router.js
```diff
@@ -12,15 +12,9 @@ import HostingEarnings from 'pages/HostingEarnings'
...
-import ScreenWidthContext from 'contexts/screenWidth'
-import styles from './PrimaryLayout.module.css' // eslint-disable-line no-unused-vars
-import 'global-styles/colors.css'
-import 'global-styles/index.css'
-
-export function PrimaryLayout () {
-  const isWide = useContext(ScreenWidthContext)
 
-  return <div styleName={cx('styles.primary-layout', { 'styles.wide': isWide }, { 'styles.narrow': !isWide })}>
+export default function HPAdminRouter () {
```
danekszy: Router file - pattern that I have mentioned back in Berlin discussions.
Single place to define our URL structure, which doesn't care about appearance.
IMHO quite readable and helpful practice ;)

We can have 2 separate routers for HPA and HF

src/components/layout/PrimaryLayout.js
danekszy: > Slot might take some arguments 

In fact we might want to consider having some params in `props.children` and then use the argument to render the footer slot. This way we would move the placement concern to the page itself (a drawback) but this would be more flexible structually.
This way we could place footer button inside of page's `<form>`

src/pages/MyProfile/MyProfile.js
danekszy: Also, header stuff will go into layout ;)

src/components/holofuel/layout/PrimaryLayout.js
```diff
@@ -0,0 +1,27 @@
+import React, { useContext } from 'react'
```
robbiecarlton: can you put this in the pod pattern of `layout/PrimaryLayout/PrimaryLayout.js`, and then export it from `layout/PrimaryLayout/index.js`, so we can keep tests and css files etc tidy

src/components/layout/PrimaryLayout.js
```diff
@@ -0,0 +1,27 @@
+import React, { useContext } from 'react'
```
robbiecarlton: see comment above about pod pattern

src/components/layout/PrimaryLayout.module.css
```diff
@@ -8,8 +8,19 @@
 
 .wide {
   padding: 50px;
+
+  /* empty */
```
robbiecarlton: remove outdated comment

src/router-holofuel.js
```diff
@@ -1,27 +1,18 @@
-import React, { useContext } from 'react'
+import React from 'react'
```
robbiecarlton: think this file should be called `HFRouter`. as a convention I think we should name files that default export a react component after that component.

I also think it would be great if this file were in a `holofuel` directory, to maintain the convention that all files for the holofuel app are in subdirectory called `holofuel`. Maybe just at the top level,

so we'd have `src/HPAdminRouter.js` and `src/holofuel/HFRouter.js`

src/router-holofuel.js
```diff
@@ -1,27 +1,18 @@
...
-import styles from './PrimaryLayout.module.css' // eslint-disable-line no-unused-vars
-import 'global-styles/holofuel/colors.css'
-import 'global-styles/holofuel/index.css'
-
-export function PrimaryLayout () {
-  const isWide = useContext(ScreenWidthContext)
-
-  return <div styleName={cx('styles.primary-layout', { 'styles.wide': isWide }, { 'styles.narrow': !isWide })}>
 
+export default function HPAdminRouter () {
```
robbiecarlton: this should be `HFRouter`

src/router.js
```diff
@@ -1,6 +1,6 @@
-import React, { useContext } from 'react'
+import React from 'react'
```
robbiecarlton: see comments about naming, above. Think this should be `src/HPAdminRouter.js`

robbiecarlton (changes requested):
This looks good. I like the PrimaryLayout/Header pattern more and more, think it's the right way to go.

Changes are mostly (maybe entirely) about filenames and directory structure.

src/router-holofuel.js
danekszy: Do we really want to go with separate directory just for this file? @robbiecarlton

robbiecarlton (approved):

PR #56, opened 2019-09-18
0699 - HoloFuel side menu

danekszy:


lorenjohnson:
@danekszy looks great! The # on the Inbox changes from the correct value of 2 to 3 and back again as I navigate between pages. Also, make it collapse if I click (tap) outside of the menu area. Otherwise, as a feature it's doing what it's supposed to and looks great. I think we should demo on this branch and you can wrap-up work on the tests and these little QA things above afterward.

lorenjohnson (approved):
👍 LGTM!

PR #57, opened 2019-09-18
1438 - Move holofuel specific files into a more sensible place.

robbiecarlton:
This simplifies the HoloFuel app setup, by moving all HoloFuel specific files into a single `holofuel` directory.

Previously there was a `holofuel` subdirectory in several places (`components`, `pages` and `global-styles`). This was dumb. (I did it so I'm allowed to say that)

This PR moves all files into a single `holofuel` directory within the main `src` directory. This `holofuel` directory has the same structure as the `src` directory (eg it contains `components`, `pages` and `global-styles` subdirectories, as well as a top level router component), just nested one level in.


lorenjohnson (approved):
👍 Great! Much prefer this.

robbiecarlton:
attn @danekszy - hopefully this will not mess up your PR, but it's a wide ranging change that includes some filename changes, so if you have any trouble with the merge, let me know and I'll help.

PR #58, opened 2019-09-23
9690 holofuel daata scripts

JettTech:


scripts/HoloFuelDataEntry/index.js
```diff
@@ -0,0 +1,310 @@
+// //////////////////////////////////////////////////////////////////////////////////// //
```
lorenjohnson, line 1: This all looks great. Curious at higher level, what it would have looked like to orchestrate this as a series of GraphQL mutations / queries to achieve the same?  Not thinking it makes more sense now, but in the future it could be worth exploring.

JettTech: Thanks for the review and thought.  Actually, the thought to call a series of GraphQL calls was one of the very discussions that @robbiecarlton and I held at the onset of this task.  We agreed on this approach in lieu of the gql for a few reasons, including cli simplicity and direct access to the conductor, but I agree that this still is a thought to consider for the integration / scenario testing approach. :+1: :bulb:

conductor-config.example.toml
```diff
@@ -105,8 +126,10 @@ id = "websocket-interface"
     [[interfaces.instances]]
     id = "holofuel"
 
-# -- HTTP admin interface (currently only used for testing the conductor is running in setup script)
+    [[interfaces.instances]]
+    id = "holofuel2"
 
+# -- HTTP admin interface (currently only used for testing the conductor is running in setup script)
 [[interfaces]]
```
zo-el, line 133: The http-interface probably needs to be set as http and not ws

holofuel-conductor-config.example.toml
```diff
@@ -0,0 +1,85 @@
...
+    [[interfaces.instances]]
+    id = "holofuel2"
+
+# -- HTTP admin interface (currently only used for testing the conductor is running in setup script)
+[[interfaces]]
+id = 'http-admin-interface'
+admin = true
+
+    [interfaces.driver]
+    type = "websocket"
```
zo-el: Here as well

zo-el (approved):
Everything else looks good. 

Can we set up a new PR to continue the second agent testing i.e. sequence testing (caused by networking??) ?

JettTech:
@zo-el - As requested, [here](https://github.com/Holo-Host/hp-admin/pull/59) is another PR from a forked branch to allow for continued testing of the second agent once networking is stable.

PR #61, opened 2019-09-24
7956 - As a HoloFuel user I can see feedback via a "flash" messages

robbiecarlton:


src/holofuel/components/FlashMessage/FlashMessage.test.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { render, fireEvent, act } from '@testing-library/react'
+import wait from 'waait'
+import FlashMessage from './FlashMessage.js'
+import useFlashMessageContext, { FlashMessageProvider } from 'holofuel/contexts/useFlashMessageContext'
+
+// Think about ways to refactor this to test fewer implementation details.
```
robbiecarlton, line 7: These tests are *fine*, but they test implementation details. Because this is a global component, avoiding that is tricky and requires basically mounting the whole app. I went for simplicity and readability of tests over purity of testing philosophy in this case.

src/holofuel/contexts/useFlashMessageContext.js
```diff
@@ -0,0 +1,19 @@
+import React, { createContext, useContext, useState } from 'react'
```
robbiecarlton, line 1: this was initially three separate files, but it ended up making sense to put them all in one place. This is a candidate for our general global state pattern, so would love for it to
a) be readable and easy to think about and use
b) be easy to set up new pieces of global state

I anticipate that this file might find a new home, and the exact structure of how we're doing this might change as the pattern matures, but this seems a fine start. 

So much so that after review I think I'll refactor screenWidth to use the same pattern.

src/index.js
```diff
@@ -23,9 +25,11 @@ export function HoloFuelApp () {
 
   return <ApolloProvider client={apolloClient}>
     <Router>
-      <ScreenWidthContext.Provider value={isWide}>
-        <HFRouter />
-      </ScreenWidthContext.Provider>
+      <HFScreenWidthContext.Provider value={isWide}>
+        <FlashMessageProvider>
```
robbiecarlton, line 29: This stack of providers at the top of the tree is my main reservation with the `context` pattern for global state at the moment. A simple solution for the source code would be to put them all in a `<Providers>` component and import that from somewhere else, but that will still leave all those components in the actual tree, which will make debugging marginally more annoying.

robbiecarlton:
@lorenjohnson - if this looks good to you, feel free to merge so it doesn't hang around til next week

src/holofuel/contexts/useFlashMessageContext.js
lorenjohnson: Makes sense.

src/index.js
lorenjohnson: Yeah, I'll watch as we go and see what emerges here. I think for now this is all fine.

lorenjohnson (approved):
LGTM 👍 We'll see this emerge more as/if we continue to use this pattern.

PR #64, opened 2019-09-26
7522 counterparty nicknames

JettTech:


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -10,6 +10,24 @@ const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
 
 const MOCK_DEADLINE = '4019-01-02T03:04:05.678901234+00:00'
 
+// Creates an array of all counterparties for a provided transaction list
+export async function getTxCounterparties (transactionList) {
```
robbiecarlton: I'm not seeing where this function is used. Is this a left over? Or am I missing something?

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -123,13 +141,22 @@ function presentTransaction (transaction) {
...
         id: result.pub_sign_key,
         nickname: result.nick
       }
+    },
+    getCounterparty: async ({ agentId }) => {
+      const result = await createZomeCall('transactions/whoami')({ agentId })
+      if (result.error) throw new Error('There was an error locating the counterparty agent nickname. ERROR: ', result.error)
+
+      return {
+        pubkey: result.pub_sign_key,
```
robbiecarlton: This should return 
```
{
  id: agentId,
  pubkey: result.pub_sign_key,
  nickname: result.nick
}
```
So that the returned `HolofuelUser` type has an id.

src/graphql-server/schema.graphql
```diff
@@ -34,6 +34,7 @@ type HostPricing {
 
 type HolofuelUser {
   id: ID
+  pubkey: String
```
robbiecarlton: why do we need `pubkey`? I'm not seeing where we're using it.

src/graphql/HolofuelCounterpartyQuery.gql
```diff
@@ -0,0 +1,6 @@
+query HolofuelCounterparty($agentId: String!) {
+  holofuelCounterparty(agentId: $agentId) {
+    pubkey
```
robbiecarlton: even if we're using `pubkey`, we should always be querying `id` as well, as that's what Apollo caching will use for identity.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -168,3 +167,12 @@ function ConfirmRejectionModal ({ transaction, handleClose, declineTransaction }
     </div>
   </Modal>
 }
+
+export function RenderNickname ({ agentId }) {
```
robbiecarlton, line 171: I feel a little uneasy about this approach, but I'm not sure that means it's wrong. 

Did you try the approach of batch fetching all the nicknames at once in Inbox component?

I can see pros and cons to either way. Let's talk about this.

robbiecarlton (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: 100% left over from the 'array' approach to resolving this useQuery (with param) call within our current component patterns.

src/graphql/HolofuelCounterpartyQuery.gql
JettTech: So the ID was explicitly removed and replaced with pubkey to prevent Apollo caching issues/conflicts arising for the need to (potentially) rewrite the HolofuelCounterparty ID returned from the new object with every new Query.  However, I agree that it isn't preferred to have a schema without an ID, as that creates an anomaly within our typing pattern.  I would be happy to pair and seek alternate solutions. :rocket:

src/graphql-server/schema.graphql
JettTech: It's used for the HolofuelCounterparty Query, instead of an ID as a current solution preventing caching concerns.

src/graphql/HolofuelCounterpartyQuery.gql
JettTech: OK. I have restored the ID to this type and we can pursue any caching issues solutions together as they arise.

PR #65, opened 2019-09-29
7513 my agent nickname

JettTech:
> Note: this includes the updates from PR #64 .  
> *Please review and merge this AFTER said PR.*


src/holofuel/components/Header/Header.js
```diff
@@ -6,20 +6,22 @@ import { withRouter } from 'react-router'
...
+  if (agentLoading) agentLoading = <h4>Loading...</h4>
+
   return <header>
     <section styleName='header'>
       <div styleName='left-nav'>
         {leftNav}
         <span styleName='title header-font'>HoloFuel</span>
       </div>
-      <div styleName='right-nav account-number header-font'>{agentId}</div>
+      <div styleName='right-nav account-number header-font'>{agent.nickname || agentLoading}</div>
```
robbiecarlton: Just looked in the ticket, and the expected behavior here is to show the last 6 of the `agent.id` until the nickname loaded, to match the behavior of the other agent nicknames.

```
{agent.nickname || presentAgentId(agent.id)}
```

src/holofuel/components/Header/Header.test.js
```diff
@@ -1,33 +1,33 @@
...
+    history
+  }
+}
 
 it('should render the title and a menu icon', () => {
   const props = {
     title: 'the title',
-    history: { push: jest.fn() }
+    history: { push: jest.fn() },
+    agent: 'AGENT NICKNAME'
```
robbiecarlton: Don't think we're use this prop.

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -19,23 +18,28 @@ export function PrimaryLayout ({
...
+  const { loading: holofuelUserLoading, data: { holofuelUser = {} } = {} } = useQuery(HolofuelUserQuery)
   const { data: { holofuelLedger: { balance: holofuelBalance } = { balance: 0 } } = {} } = useQuery(HolofuelLedgerQuery)
 
   const inboxCount = transactions.length
-  const agentId = presentAgentId(holofuelUser.id)
+
+  let agent = {}
+  let agentLoading
+  if (holofuelUserLoading) agentLoading = true
+  else agent = { ...holofuelUser }
```
robbiecarlton: this doesn't need to be conditional. You can always assign `agent` to `holofuelUser` here, because `holofuelUser` will be empty by default.

In fact, I'm pretty sure lines 26 - 29 can be removed if you change line 37 to
`<Header {...headerProps} agent={holofuelUser} agentLoading={holofuelUserLoading} hamburgerClick={hamburgerClick} />`

and line 42 to
`agentLoading={holofuelUserLoading}`

robbiecarlton (changes requested):
Just looking at the pivotal ticket and it says no need to use a loading indicator for this feature and I think that's right. As we already have the agentId, I think it's fine to just replace that when the loading has finished.

PR #67, opened 2019-09-29
5400 copy holofueluser and holofuelcounterparty IDs

JettTech:
> NB: this includes the updates from PR #66, #65, #64, and #61 .  
> *Please review and merge this AFTER said PRs.*

PR #66, opened 2019-09-29
7572 request/offer counterparty names

JettTech:
- PR Abides by all Pivotal Acceptance Criteria for ticket #7572; however, there are known interaction issues with the Flash Message component/context ( PR #61 ).  Consequently, error messages are currently displayed inline on screen, instead of through the flash message context.

> NB: this includes the updates from PR #65, #64, and PR #61 .  
> *Please review and merge this AFTER said PRs.*

PR #68, opened 2019-10-01
7279 pay confirmation modal

JettTech:
> NB: this includes the updates from PR #67, #66, #65, and #64
> *Please review and merge this AFTER said PRs.*

### DIFF
--- a/src/holofuel/pages/Inbox/Inbox.js
+++ b/src/holofuel/pages/Inbox/Inbox.js
@@ -1,6 +1,5 @@
 import React, { useState } from 'react'
 import cx from 'classnames'
-import _ from 'lodash'
 import { isEmpty } from 'lodash/fp'
 import { useQuery, useMutation } from '@apollo/react-hooks'
 import HolofuelCounterpartyQuery from 'graphql/HolofuelCounterpartyQuery.gql'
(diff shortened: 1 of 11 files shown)


src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -109,51 +123,59 @@ function AcceptButton ({ transaction: { id } }) {
...
-    refetchQueries: [{
-      query: HolofuelActionableTransactionsQuery
-    }]
-  })
-}
-
-function PayButton ({ transaction: { id, amount, counterparty } }) {
-  const pay = useOffer(id, amount, counterparty)
+function PayButton ({ showConfirmationModal, transaction }) {
+  // const pay = useOffer(id, amount, counterparty)
```
robbiecarlton: remove comment

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -109,51 +123,59 @@ function AcceptButton ({ transaction: { id } }) {
...
+
   const onYes = () => {
-    declineTransaction(id)
+    actionHook(actionParams)
     handleClose()
   }
 
   return <Modal
-    contentLabel={`Reject ${type}?`}
+    contentLabel={`${_.capitalize(action)} ${type}?`} // WAS: contentLabel={`Reject ${type}?`}
```
robbiecarlton: I think it would be more readable just to set a `contentLabel` var in each of the two cases. So this line would just read `contentLabel={contentLabel}`, and that value is set above based on the case.

robbiecarlton (changes requested):

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -145,6 +144,7 @@ export function ConfirmationModal ({ transaction, handleClose, declineTransactio
   const { id, counterparty, amount, type } = transaction
 
   const action = type === TYPE.request ? 'pay' : 'decline'
+  const contentLabel = `${_.capitalize(action)} ${type}?`
```
robbiecarlton: I meant that we should set this in the switch statement below, and not capitalize the action. So there would be two cases
```
contentLabel = 'Pay request'
```
and
```
contentLabel = `Reject ${type}?`
```

robbiecarlton (changes requested):

src/holofuel/pages/Inbox/Inbox.js
JettTech: No problem; this is an easy update.  Why recommend this approach?

robbiecarlton: `'${_.capitalize(action)} ${type}?'` this requires a bunch of thinking to read, and doesn't really buy us anything.

robbiecarlton: @JettTech that's my suggestion, and I'm happy to defer to you. If you want to merge without changing this, go ahead.

robbiecarlton (approved):

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -169,19 +194,19 @@ function ConfirmRejectionModal ({ transaction, handleClose, declineTransaction }
...
     </CopyAgentId>
-  }
-  return <CopyAgentId agent={data.holofuelCounterparty}>
-    {data.holofuelCounterparty.nickname}
-  </CopyAgentId>
+  } else if (holofuelCounterparty.nickname && copyId) {
+    return <CopyAgentId agent={holofuelCounterparty}>
+      {holofuelCounterparty.nickname}
+    </CopyAgentId>
+  } else return <>{holofuelCounterparty.nickname}</>
```
robbiecarlton, line 211: still think i prefer the approach of just wrapping this component in a <CopyAgentId> when we need to, rather than trying to build the functionality in here. This kind of compound if statement is harder to read and think about, and error prone.

robbiecarlton (changes requested):

robbiecarlton:
one last change, otherwise looks good 👍

src/holofuel/pages/Inbox/Inbox.js
JettTech: The reason I wrap the component here is because we don't have access to the holofuelConterparty nickname outside of the RenderNickname Component, which CopyAgentId needs as a value for it's `agent` prop.

Open to other ideas though. :+1:

robbiecarlton (approved):

PR #70, opened 2019-10-03
4119 - Show flash message on creating offer

robbiecarlton:


src/holofuel/pages/CreateRequest/CreateRequest.js
```diff
@@ -110,10 +117,7 @@ export function RenderNickname ({ agentId, setCounterpartyNick }) {
     </>
   }
 
-  if (error || !data.holofuelCounterparty.nickname) {
-    return <>No nickname available.</>
-  } else {
-    setCounterpartyNick(data.holofuelCounterparty.nickname)
```
robbiecarlton, line 116: we don't want to run functions with side effects in a render method. They should always be wrapped in a `useEffect` hook. 
`useEffect` is the functional component equivalent of `componentDidMount` and `componentDidUpdate` in a class component.

JettTech (approved):
Good note about the hooks within the render functions.  All good great. :+1:

PR #72, opened 2019-10-04
9692 - Create mock for HTTP REST interface (for one HPOS HTTP endpoint)

robbiecarlton:


robbiecarlton:
The specific endpoint is a little bit of guesswork from me (and is almost certainly wrong. Doubtful that we can update the os-version). We don't yet have API definition from holo-central yet. But the infrastructure is here and it will be trivial to adjust to the real endpoint once we have the API definition.

JettTech (approved):
Nice work.  The only thing I would request, is to add a test for this endpoint into the resolvers test, as a general test of the setup.

robbiecarlton:
> Nice work. The only thing I would request, is to add a test for this endpoint into the resolvers test, as a general test of the setup.

Yup, good call

PR #73, opened 2019-10-07
Nix setup

zo-el:
### Tasks:
- [x] auto-generate conductor on nix-shell startup
- [x] Spins up conductor when you enter `nix-shell`
- [x] `hp-test` nix script to run the unit tests

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -29,6 +29,11 @@ let
     };
   }) config.agents;
 
+
+  multiInterfaceInstanceConfig = dna: map (agent: {
+    id = dna.name+"::"+agent.id;
+  }) config.agents;
+
   flatten = x:
     if builtins.isList x
     then builtins.concatMap (y: flatten y) x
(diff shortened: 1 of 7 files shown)


default.nix
```diff
@@ -0,0 +1,119 @@
...
+
+with pkgs;
+
+let
+
+  config = import ./config.nix;
+
+  name = "hp-test";
+
+  script = pkgs.writeShellScriptBin name
```
peeech: This line creates a command in nix-shell "hp-test", but you are not using it in shellHook. I think that if you're trying to expose it in nix-shell it might be nice to write in README.md sth like "type `hp-test` to run tests". Or if there is no plan to use it I think deleting this part might make sense.

default.nix
```diff
@@ -0,0 +1,119 @@
+{ pkgs ? import ./pkgs.nix {} }:
+
+with pkgs;
+
+let
+
+  config = import ./config.nix;
```
peeech, line 6: Is there a way to gracefully exit if the config.nix is not present? @yegortimoshenko

peeech (commented):
LGTM overall.

example.config.nix
```diff
@@ -0,0 +1,22 @@
...
+    keystore_file = "<Path to keystore file>";
+    public_address = "<public key>";
+  }
+  {
+    id = "agent-2-id";
+    name = "Agent 2";
+    keystore_file = "<Path to keystore file>";
+    public_address = "<public key>";
+  }];
+
```
JettTech, line 21: I like this option to customize or add additional agents.  :rocket:

JettTech: Also, you had mentioned your intention to have nix build a default agent in addition to this agent config.  I think this is a great idea and would love to see this included in this PR.

zo-el: We can update this in the nixification process

JettTech (commented):
Overall, good work! :+1: I' think it's ready for a merge.

PR #77, opened 2019-10-08
5680 holofuel data layer update

JettTech:


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -92,32 +92,32 @@ function presentPendingOffer (transaction) {
...
       return presentRequest({ origin, event, stateDirection, eventTimestamp: timestamp.event, fees: parsedAdjustment.fees })
     }
-    // NOTE: 'approved' only indicates that a payment was offered (could be in response to a request or an isolate payment)
+    // 'approved' only indicates that a payment was offered (could be in response to a request or an isolate payment)
     case 'approved': {
       return presentOffer({ origin, event, stateDirection, eventTimestamp: timestamp.event, fees: parsedAdjustment.fees })
     }
     default:
-      throw new Error('Error: No transaction stateState was matched. Current transaction stateStage : ', stateStage)
+      throw new Error('Error: No transaction stateStag  e was matched. Current transaction stateStage : ', stateStage)
```
zo-el:
```suggestion
      throw new Error('Error: No transaction stateStage was matched. Current transaction stateStage : ', stateStage)
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -232,10 +251,17 @@ const HoloFuelDnaInterface = {
 
     accept: async (transactionId) => {
       const transaction = await HoloFuelDnaInterface.transactions.getPending(transactionId)
-      await createZomeCall('transactions/receive_payments_pending')({ promises: transactionId })
+      const result = await createZomeCall('transactions/receive_payments_pending')({ promises: transactionId })
+
+      if (Object.entries(result).length > 1) throw new Error('Multiple accepted offers were returned, when only one was triggered. \n Here is the Tansaction Origin ID of requested accept payment, and the resulting payments accepted : ', transactionId, Object.entries(result))
```
robbiecarlton: Hmm. This feels a little fishy to me. Is this something the dna does? Is it a bug? If so, can we create an issue for the dna code? If not a bug, we should handle multiple offers returned (maybe filter by transactionId)

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -232,10 +251,17 @@ const HoloFuelDnaInterface = {
...
+
+      const acceptedPaymentHash = Object.entries(result)[0][1]
+      if (acceptedPaymentHash.Err) throw new Error('There was an error accepting the payment for the referenced transaction. ERROR: ', acceptedPaymentHash.Err)
+      console.log('RECEIVE_PAYMENTS_PENDING SUCCESS HASH : ', acceptedPaymentHash.Ok)
```

src/mock-dnas/holofuel.js
```diff
@@ -332,17 +297,28 @@ function listPending ({ origins }) {
   throw new Error('Array value for origins param of list_pending is not supported in the mock dna')
 }
 
+function receivedPaymentsHashMap (promises) {
+  return promises.forEach(promise => {
```
robbiecarlton: `.forEach` always returns `undefined`, so this function always returns `undefined`. `.map` will return an array that is the result of applying the function to each item in the original array. But that will still do something a little strange in this case. If you used `.map` the result of this function would be an array of objects, each of which has exactly one key.

`receivedPaymentsHashMap([1, 2, 3])` 
->
```
[{
  1: {Ok: { result: /* hashed string of '1accepted' */} }
},
{
  2: {Ok: { result: /* hashed string of '2accepted' */} }
}, 
{
  3: {Ok: { result: /* hashed string of '3accepted' */} }
}]
```

But perhaps that's what you want?

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -172,7 +191,7 @@ const HoloFuelDnaInterface = {
       const noDuplicateIds = _.uniqBy(listOfNonActionableTransactions, 'id')
       return noDuplicateIds.filter(tx => tx.status === 'pending').sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
     },
-    getPending: async transactionId => {
+    getPending: async (transactionId) => {
```
robbiecarlton, line 210: My vote would be we don't wrap single params in parens in function literals.

so eg

`(firstParam, secondParam) => /* do stuff with firstParam and secondParam */`

but

`onlyParam => /* do stuff with onlyParam */`

What do you think about that convention?

robbiecarlton (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
zo-el: I vote that it should be in parens. 
I feel it makes it more readable. 
But I don't have any strong feelings about it.

zo-el (approved):
:+1:

robbiecarlton (approved):

robbiecarlton:
LGTM

PR #79, opened 2019-10-11
use renderAndWait in inbox.test

robbiecarlton:
@JettTech  - What do you think? I think it improves readability of the tests considerably. I just did it in `inbox.test`, but there's a chore to do it everywhere else.

### DIFF
--- a/src/utils/test-utils.js
+++ b/src/utils/test-utils.js
@@ -1,0 +1,11 @@
+import { render, act } from '@testing-library/react'
+import wait from 'waait'
+
+export async function renderAndWait (ui, ms = 0, options = {}) {
+  let queries
+  await act(async () => {
+    queries = render(ui, options)
+    await wait(ms)
+  })
+  return queries
+}
--- a/src/holofuel/pages/Inbox/Inbox.test.js
+++ b/src/holofuel/pages/Inbox/Inbox.test.js
@@ -1,6 +1,5 @@
 import React from 'react'
-import Modal from 'react-modal'
-import { render, fireEvent, act, within } from '@testing-library/react'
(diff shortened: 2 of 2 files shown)


src/utils/test-utils.js
```diff
@@ -0,0 +1,11 @@
...
+import wait from 'waait'
+
+export async function renderAndWait (ui, ms = 0, options = {}) {
+  let queries
+  await act(async () => {
+    queries = render(ui, options)
+    await wait(ms)
+  })
+  return queries
+}
```
JettTech, line 11: Great idea and nicely done!  :100:

JettTech (approved):
LOVE this! :rocket:  What you did here with the new `renderAndWait` helper function is so simple, but SO impactful, and dramatically improves ease of reading. 

This also highlights a great point.  I think were at a good moment now, where we can go through our code base as it stands currently and agree on even more functional extractions / improvements to implement.  Shall we add a backlog item?  Thoughts?

robbiecarlton:
> LOVE this! 🚀 What you did here with the new `renderAndWait` helper function is so simple, but SO impactful, and dramatically improves ease of reading.
> 
> This also highlights a great point. I think were at a good moment now, where we can go through our code base as it stands currently and agree on even more functional extractions / improvements to implement. Shall we add a backlog item? Thoughts?

Great. Glad you like it. This PR is failing linting and I haven't figured out why yet so not going to merge until I do.

Also, we're no longer getting the reports from CI on our PRs which I haven't looked into yet but is concerning.

I think putting in a backlog item for this change makes sense. Did you also mean a backlog item for general code review/refactoring?

PR #80, opened 2019-10-11
5115 counterparty batch loading

JettTech:


package.json
```diff
@@ -6,6 +6,7 @@
     "start": "rm -rf ./node_modules/.cache/babel-loader && node scripts/start.js",
     "start:holofuel": "REACT_APP_HOLOFUEL_APP=true npm start",
     "build": "node scripts/build.js",
+    "build:holofuel": "REACT_APP_HOLOFUEL_APP=true npm run build",
```
robbiecarlton: does this work? And why is it in this PR?

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -10,6 +10,25 @@ const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
 
 const MOCK_DEADLINE = '4019-01-02T03:04:05.678901234+00:00'
 
+// Creates an array of all counterparties for a provided transaction list
+export async function getTxCounterparties (transactionList) {
+  const counterpartyList = transactionList.map(tx => {
+    const { counterparty } = tx
+    return counterparty
+  })
```
robbiecarlton: This can be simplified to
```
const counterpartyList = transactionList.map(({ counterparty }) => counterparty)
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -10,6 +10,25 @@ const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
...
+  const agentDetailsList = []
+  for (let i = 0; i < counterpartyList.length; i++) {
+    const agentId = counterpartyList[i]
+    const agent = await new Promise(resolve => {
+      const findAgent = HoloFuelDnaInterface.user.getCounterparty({ agentId })
+      resolve(findAgent)
+    })
```
robbiecarlton: I believe this can be simplified to
```
const agent = await HoloFuelDnaInterface.user.getCounterparty({ agentId })
```

robbiecarlton: but there's a function in utils that makes this even simpler. I think you can replace lines 19-27 with

```
import { promiseMap } from 'utils'

...

const agentDetailsList = await promiseMap(counterpartyList, agentId => HoloFuelDnaInterface.user.getCounterparty({ agentId }))
```

src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -94,16 +107,24 @@ const TransactionTableHeading = ({ content }) => {
   </th>
 }
 
-export function TransactionRow ({ transaction, showCancellationModal, completed }) {
+export function TransactionRow ({ transaction, showCancellationModal, completed, counterpartyList }) {
   const { id, timestamp, amount, counterparty, direction, fees, presentBalance, notes } = transaction
+
+  let counterpartyNick = 'Loading...'
+  if (counterpartyList) counterpartyNick = counterpartyList.find(agent => agent.id === counterparty).nickname
```
robbiecarlton: I think there's a way we can avoid this using Apollo Cache. We can have `useFetchCounterparties` update the cache so that the nickname is in `Transaction.user`. I just looked in `Schema.graphql` and currently `Transaction.counterparty` is of type String, and there's no field of type `HolofuelUser` on `Transaction`. I think part of this will be adding that field.

robbiecarlton (changes requested):

package.json
JettTech: Good catch; somehow this build script test was maintained when creating a new branch.  This build approach *does* work btw, but I did not mean for this branch to include that change. I'll remove it.

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Sure can. Updating now. :+1:

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: So, at first, I did actually write `const agent = await HoloFuelDnaInterface.user.getCounterparty({ agentId })`.  

However, because we're calling await within an iterator and the return value of async functions is always a Promise, `agentDetailsList` will actually just return an Array of unresolved promises.  Consequently, the value for each iteration must be resolved before we can make use of the resulting Array, which is why each call to `HoloFuelDnaInterface.user.getCounterparty({ agentId })` is wrapped in a Promise.

JettTech: Another option is to wrap a map function of the `counterpartyList` array in a Promise.all(); however, from my experience, this approach requires a bit more setup / code to attain the same outcome.

JettTech: Open to new ideas or additional thoughts though. :+1:

src/holofuel/pages/TransactionHistory/TransactionHistory.js
JettTech: Interesting idea. I'l look into this. :+1:

JettTech: Done.

.gitignore
```diff
@@ -20,8 +20,11 @@
 .env.production.local
 tmp-storage
 tmp
-conductor-config.toml
 holofuel-conductor-config.toml
+conductor-config.toml
+conductor.pid
+conductor.log
+config.nix
```
robbiecarlton: this has since been removed from  gitignore, so make sure to take it out here so it doesn't put it back in.

src/data-interfaces/HoloFuelDnaInterface.js
robbiecarlton: > However, because we're calling await within an iterator and the return value of async functions is always a Promise, agentDetailsList will actually just return an Array of unresolved promises. Consequently, the value for each iteration must be resolved before we can make use of the resulting Array, which is why each call to HoloFuelDnaInterface.user.getCounterparty({ agentId }) is wrapped in a Promise.

This is what `promiseMap` from `utils` addresses. It maps a promise returning function over an array and returns  a single promise that resolves to the list of resolved results (it's more intuitive than that sounds).

Did you try my suggestion above using `promiseMap`? It should work.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -35,10 +36,41 @@ function useDecline () {
...
+    })
+  }
+
+  let response
+  if (loading) response = { loading: true }
+  if (error) response = { error: `Error: ${error}` }
+  else response = holofuelInboxCounterparties
+  return response
+}
+
```
robbiecarlton: This is an interesting approach, and still not the one I meant by "using the cache". I think the easiest thing is for us to pair on this for 20 minutes. I'm not sure my approach will end up being better but I'd still like to see it implemented so we can compare.

robbiecarlton (changes requested):
I think it's going to be easier for us to pair on this rather than going back and forth in comments. How's that sound?

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Got it. Just read through `promiseMap`, and it will do the trick.  Imported and implemented. :+1:

src/holofuel/pages/Inbox/Inbox.js
JettTech: Happy to share ideas and compare options. :100:

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -10,11 +11,37 @@ const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
...
-const presentRequest = ({ origin, event, stateDirection, eventTimestamp, counterparty, amount, notes, fees }) => {
+// Creates an array of all counterparties for a provided transaction list
+export async function getTxCounterparties (transactionList) {
+  const counterpartyList = transactionList.map(({ counterparty }) => counterparty.id)
+  const agentDetailsList = await promiseMap(counterpartyList, agentId => HoloFuelDnaInterface.user.getCounterparty({ agentId }))
+  const noDuplicatesAgentList = _.uniqBy(agentDetailsList, 'id')
+  return noDuplicatesAgentList
+}
+
+// // Creates an array of all counterparties for a provided transaction list
```
robbiecarlton: remove commented code

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -35,13 +35,46 @@ function useDecline () {
...
 export default function Inbox () {
   const { data: { holofuelActionableTransactions: transactions = [] } = {} } = useQuery(HolofuelActionableTransactionsQuery)
+
   const payTransaction = useOffer()
   const declineTransaction = useDecline()
   const [modalTransaction, setModalTransaction] = useState()
   const isTransactionsEmpty = isEmpty(transactions)
 
+  // const counterpartyList = useFetchCounterparties()
```
robbiecarlton: remove commented code

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -35,13 +35,46 @@ function useDecline () {
...
+  useFetchCounterparties()
```
robbiecarlton: nitpicking, but I'd move this to line 69 so that it's next to the code it relates to

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -35,13 +35,46 @@ function useDecline () {
...
+      }
+    })
+  }
+
+  let response
+  if (loading) response = { loading: true }
+  else if (error) response = { error: `Error: ${error}` }
+  else response = holofuelInboxCounterparties
+  return response
+}
```
robbiecarlton: given that we don't use this return value, let's omit it, and maybe put a comment in its place `// this function updates the cache of other Queries, we don't use a return value directly`

src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -27,12 +26,55 @@ function useCancel () {
...
+
 // Display - Functional Components with Hooks :
 export default function TransactionsHistory () {
   const { data: { holofuelCompletedTransactions: completedTransactions = [] } = {} } = useQuery(HolofuelCompletedTransactionsQuery)
   const { data: { holofuelWaitingTransactions: pendingTransactions = [] } = {} } = useQuery(HolofuelWaitingTransactionsQuery)
 
   const cancelTransaction = useCancel()
+
+  useFetchCounterparties()
+  // const counterpartyList = useFetchCounterparties()
```
robbiecarlton: remove commented code and move this up to the top with the other queries as in Inbox

src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -27,12 +26,55 @@ function useCancel () {
...
+      }
+    })
+  }
+
+  let response
+  if (loading) response = { loading: true }
+  else if (error) response = { error: `Error: ${error}` }
+  else response = holofuelHistoryCounterparties
+  return response
+}
```
robbiecarlton: remove return value and add comment as in Inbox

src/holofuel/pages/TransactionHistory/TransactionHistory.test.js
```diff
@@ -79,15 +79,15 @@ describe('TransactionsHistory', () => {
...
+            const whois = await HoloFuelDnaInterface.user.getCounterparty({ agentId: fullRowContent[rowIndex].counterparty.id })
             const notesDisplay = transaction.notes === null ? 'none' : transaction.notes
             const dateDisplay = presentDateAndTime(transaction.timestamp).date
             const timeDisplay = presentDateAndTime(transaction.timestamp).time
             const amountToMatch = transaction.direction === 'outgoing' ? `(${presentHolofuelAmount(transaction.amount)})` : presentHolofuelAmount(transaction.amount)
 
             expect(within(getByTestId('cell-date')).getByText(dateDisplay)).toBeInTheDocument()
             expect(within(getByTestId('cell-time')).getByText(timeDisplay)).toBeInTheDocument()
-            expect(within(getByTestId('cell-counterparty')).getByText(whois.nickname) || within(getByTestId('cell-counterparty')).getByText(presentAgentId(transaction.counterparty))).toBeInTheDocument()
+            expect(within(getByTestId('cell-counterparty')).getByText(whois.nickname) || within(getByTestId('cell-counterparty')).getByText(presentAgentId(transaction.counterparty.id))).toBeInTheDocument()
```
robbiecarlton: let's be specific about what we're testing here. Can we just test for `nickname` and not for `id`?

src/holofuel/pages/TransactionHistory/TransactionHistory.test.js
```diff
@@ -251,7 +243,7 @@ describe('TransactionsHistory', () => {
       const row = getAllByRole('row')
       const { getByTestId } = within(row[1])
       expect(rowContent.status).toBe('completed')
-      expect(within(getByTestId('cell-counterparty')).getByText(presentAgentId(rowContent.counterparty)) || within(getByTestId('cell-counterparty')).getByText(whois.nickname)).toBeInTheDocument()
+      expect(within(getByTestId('cell-counterparty')).getByText(presentAgentId(rowContent.counterparty.id)) || within(getByTestId('cell-counterparty')).getByText(whois.nickname)).toBeInTheDocument()
```
robbiecarlton: same, just test for `nickname`

src/holofuel/pages/Inbox/Inbox.test.js
```diff
@@ -5,17 +5,16 @@ import { ApolloProvider } from '@apollo/react-hooks'
 import { MockedProvider } from '@apollo/react-testing'
 import moment from 'moment'
 import apolloClient from 'apolloClient'
-import Inbox, { TransactionRow, RenderNickname } from './Inbox'
+import Inbox, { TransactionRow } from './Inbox'
```
robbiecarlton, line 8: in this file there is one test (on line 40) that the nicknames are rendered, but it's currently commented out. Can you uncomment that part and make sure the tests are testing that nicknames are rendering

robbiecarlton (changes requested):
Some small changes but the approach is good! How do you feel about this version?

robbiecarlton (approved):
LGTM

PR #88, opened 2019-10-21
0215 settings w version tracker

JettTech:


JettTech:
The changes within this PR are now reflected in PR #90

PR #90, opened 2019-10-22
0215 0.1.2 hpadmin settings page

JettTech:


src/mock-dnas/hpos.js
```diff
@@ -63,6 +63,7 @@ const data = {
     // Forces HoloPortOS upgrade. - Returns Status Code ONLY (200 if OK/Successful, 400 if Bad Request, or 401 if Unauthorized).
     // TODO: Talk to HCen to see if hydra/hpos will return new Status Obj upon successful update.
     '/v1/upgrade': () => {
+      console.log('Code 200: You have now updated your HPOS Version.')
```
robbiecarlton, line 66: Did you mean to leave this log in? I feel ambivalent about it, leaning towards removing it. We already have logs that tell us which dna endpoints are being called.

src/pages/Settings/Settings.js
```diff
@@ -1,146 +1,182 @@
...
+import Modal from 'components/Modal'
 import PrimaryLayout from 'components/layout/PrimaryLayout'
 import Button from 'components/Button'
-import Input from 'components/Input'
+import HposSettingsQuery from 'graphql/HposSettingsQuery.gql'
+import HposStatusQuery from 'graphql/HposStatusQuery.gql'
+import HposUpdateVersionMutation from 'graphql/HposUpdateVersionMutation.gql'
 
-import './Settings.module.css'
+export const createLabelfromSnakeCase = string => string.replace(/([A-Z])/g, ' $1').split(' ').map(word => capitalize(word)).join(' ')
```
robbiecarlton: This is clever, but I think the code would be more self documenting if this were just a dictionary lookup
```
getLabelFromPortName = portname => ({
  primaryPort: 'Primary Port'.
  // ...etc
}[portname])
```

Does that make sense? Maybe I'm missing something.

src/pages/Settings/Settings.js
```diff
@@ -1,146 +1,182 @@
...
-  toggleSshAccess: () => Promise.resolve(true)
+export function UpdateSoftwareModal ({ availableVersion, updateVersion, handleClose, settings }) {
+  if (!availableVersion) return null
+  const onYes = () => {
+    updateVersion(availableVersion)
+    handleClose()
+  }
+  return <Modal
+    contentLabel={`Update ${settings.deviceName}?`}
+    isOpen={!!availableVersion}
```
robbiecarlton: I think overloading `availableVersion` like this makes the code harder to understand. Is it really saving us much to avoid just having a separate `modalIsOpen` state in the parent?

src/pages/Settings/Settings.js
```diff
@@ -1,146 +1,182 @@
...
+  const [softwareUpdateVersion, setSoftwareUpdateVersion] = useState()
+  const showSoftwareUpdateModal = availableVersion => setSoftwareUpdateVersion(availableVersion)
 
-  const onSubmit = settings => {
-    updateSettings(settings)
+  let updateAvailable = false
+  let availableVersion, currentVersion
+  if (!isEmpty(status) && !isEmpty(status.versionInfo)) {
+    availableVersion = status.versionInfo.availableVersion
+    currentVersion = status.versionInfo.currentVersion
```
robbiecarlton: you can use `lodash/fp/get` for this
```
const availableVersion = get('versionInfo.availableVersion', status)
```

src/pages/Settings/Settings.js
```diff
@@ -1,146 +1,182 @@
...
   }
+  if (!isEmpty(availableVersion) && !isEmpty(currentVersion) && (availableVersion !== currentVersion)) updateAvailable = true
```
robbiecarlton: `const updateAvailable = (!isEmpty(availableVersion) && !isEmpty(currentVersion) && (availableVersion !== currentVersion))`

avoids having to predefine  `updateAvailable` on line 35

src/pages/Settings/Settings.js
```diff
@@ -1,146 +1,182 @@
...
-        errors={errors} />
-
-      <Button variant='primary' wide name='update-settings' value='Submit'>Save Changes</Button>
-    </form>
+    <header styleName='jumbotron-header'>
+      {!isEmpty(settings) && <>
+        <HashAvatar seed={settings.hostPubKey} styleName='avatar-image' />
+        <h2> {settings.hostName ? `${settings.hostName}'s` : 'Your'} HoloPort </h2>
+      </> }
+      {/* TODO: Find out what the below number should link to... */}
```
robbiecarlton, line 49: is this captured in pivotal?

src/pages/Settings/Settings.js
```diff
@@ -1,146 +1,182 @@
...
+    </thead>
+    <tbody>
+      {children}
+    </tbody>
+  </table>
+}
 
-    <Button name='factory-reset' variant='danger' wide onClick={() => push('/factory-reset')}>Factory Reset</Button>
-  </PrimaryLayout>
+export function SettingsRow ({ label, content, showSoftwareUpdateModal, updateAvailable, versionTable }) {
```
robbiecarlton: I think I would prefer just having a separate component `SoftwareUpdateRow` that uses the same styles, rather than trying to overload this component. We get a little more duplication that way but it's pretty invisible, and makes the two components much more readable and easier to think about it.

src/pages/Settings/Settings.js
```diff
@@ -1,146 +1,182 @@
...
-  ...inputProps
-}) {
-  return <>
-    {label && <label styleName='settingsLabel' htmlFor={name}>{label}</label>}
-    <Input name={name} id={name} type={type} placeholder={label} ref={register} {...inputProps} />
-    {errors[name] && <small styleName='field-error'>{errors[name].message}</small>}
-  </>
+function SoftwareUpdateButton ({ availableVersion, showSoftwareUpdateModal }) {
+  return <Button
+    onClick={() => showSoftwareUpdateModal(availableVersion)}
```
robbiecarlton: yeah, I think it's a smell that we need SoftwareUpdateButton to know `availableVersion`

line 146 should just read
```
onClick={showSoftwareUpdateModal}
```
or at worst, 
```
onClick={() => showSoftwareUpdateModal(true)}
```

robbiecarlton (changes requested):

src/pages/Settings/Settings.js
```diff
@@ -1,146 +1,182 @@
...
+        <SettingsRow
+          label='Network ID'
+          // TODO : Determine desired approach for diplaying full networkId Hash...
+          content={!isEmpty(status) && status.networkId ? presentHash(status.networkId, 14) : NOT_AVAILABLE} />
+      </SettingsTable>
+
+      <SettingsTable header='Access Port Numbers'>
+        {!isEmpty(status) && !isEmpty(status.ports)
+          ? Object.entries(status.ports).map(port => {
+            if (port[0] === '__typename') return
```
robbiecarlton: This returns `undefined` which will upset `React` (think this is why you're getting a warning here in the console). `return null` instead when you're returning something to be rendered by `React`.

robbiecarlton (changes requested):

src/pages/Settings/Settings.js
JettTech: This works too and upon second reflection, I believe would also be better control over the output. 
 I'll switch to this approach.

src/pages/Settings/Settings.js
JettTech: No problem; this is an easy update. 

For clarity though, what is the reason for this preference?  As far as I can tell from the lodash docs, there is no additional error handling provided by this method (although there is a default state option for unresolved values that we could use). What added value does the lodash get method provide for us here?

src/pages/Settings/Settings.js
JettTech: Note is now both in the wireframes and Pivotal.

src/mock-dnas/hpos.js
JettTech: I could go either way, but do have a slight preference to leave it in as a quick visual/semantic reminder.

src/pages/Settings/Settings.js
robbiecarlton: get doesn't throw if the attributes aren't present, it just returns a value (default `undefined` I think). So you can avoid the `isEmpty` checks, so

```
const availableVersion = get('versionInfo.availableVersion', status)
```

replaces lines 37-39

JettTech: Got it. :+1:

robbiecarlton (approved):

robbiecarlton:
Just fix the merge conflict and this is good to go

PR #91, opened 2019-10-23
Wasm crypto

willemolding:
# holo-crypto WASM

First pass at a reusable lib that implements the required crypto algorithms required by Holo. This contains two crates, one to be consumed by Rust and another that builds a WASM binary plus javascript wrappers. Based largely on https://github.com/Holo-Host/hclient2/pull/1/ but changes up the API and uses a bunch of tricks to shrink the WASM size.

## Consuming the WASM lib

The library can be installed and imported by other javascript modules in the project by first building the lib by running `wasm-pack build` in the holo-crypto-wasm dir and then adding the following to a package.json
```json
"dependencies": {
    ...
    "holo-crypto-wasm": "file:../holo-crypto/crates/holo-crypto-wasm/pkg",
}
``` 
and then imported as normal in js/ts
```javascript
import { AdminKeyPair, WebUserKeyPair } from "holo-crypto-wasm";
```

## API

The lib exposes a very simple object style API to javascript. A KeyPair (either admin or webuser) can be created by instantiating the appropriate object and passing a passphrase, email, salt etc. The function sign() can then be called on this object to sign a string payload. The public and private keys never leave the WASM memory space. 

```javascript
import { AdminKeyPair } from "holo-crypto-wasm";
const kp = new AdminKeyPair(
      "some@email.test",
      "p4ssw0rd",
      "HCA01adefasdtbeodbgahfawnf"
    )
const sig = kp.sign("some message")
```

Both these functions will throw Javascript exceptions with human readable error information if the Rust code errors.

## Crypto Considerations

This crate makes use of the following implementations 

*https://github.com/FauxFaux/argon2min* is used for password hashing. This implements the Argon2 algorithm and variations in pure Rust. This library appears to not be particularly active and its audit status is unknown. 

*https://github.com/dalek-cryptography/ed25519-dalek* is used for public key pair generation and signing also in pure Rust. These guys are super legit https://dalek.rs/. This project is active and used by various other projects in the crypto space (Monero, Solana). It appears this has been through [an audit](https://blog.quarkslab.com/security-audit-of-dalek-libraries.html) but this probably needs to be looked at further. 

## WASM sizes and Benchmarking

The holo-crypto-wasm crate is able to build to a WASM binary that is about 180kB with some optimizations enabled. This is certainly small enough to load client side without much trouble This possibly could be reduced further with aggressive optimization. 

---

Some trivial benchmarks can be obtained by running the included benchmarks page in the browser. 

function | avg time (ms)
---------- | -----------------
Keypair.new() | 304.3
keypair.sign() | 0.2

This suggests both the key generation and signing are probably fast enough for production use


src/holo-crypto/crates/holo-crypto-wasm/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+#[wasm_bindgen]
+pub struct AdminKeyPair(KeyPair);
+
+#[wasm_bindgen]
+impl AdminKeyPair {
+    #[wasm_bindgen(constructor)]
+	pub fn new(email: &str, pass: &str, hc_pub_key: &str) -> Result<AdminKeyPair, JsValue> {
+		let combined_passphrase = pass.to_owned() + email;
```
willemolding: I was not sure how we want to handle the email, password and hc_pub_key. For the first pass I am concatenating the password and email and using the hc_pub_key as salt. This probably needs some more consideration

src/holo-crypto/crates/holo-crypto-wasm/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+
+
+#[wasm_bindgen]
+pub struct WebUserKeyPair(KeyPair);
+
+#[wasm_bindgen]
+impl WebUserKeyPair {
+    #[wasm_bindgen(constructor)]
+	pub fn new(email: &str, pass: &str, salt: &str, happ_hash: &str) -> Result<WebUserKeyPair, JsValue> {
+		let combined_passphrase = pass.to_owned() + email + happ_hash;
```
willemolding: similarly here

src/holo-crypto/crates/holo-crypto/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+crate-type = ["lib"]
+
+[dependencies]
+base64 = "0.10.1"
+ed25519-dalek = "1.0.0-pre.2"
+wasm-bindgen = "0.2"
+
+[dependencies.argon2min]
+git = "https://github.com/Holo-Host/argon2min"
+branch = "2019-10-13-holo-config"
```
willemolding, line 17: I mentioned in mattermost as well but @yegortimoshenko what was your reasoning around forking this repo and did you make any important changes?

src/holo-crypto/crates/holo-crypto/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+        // threads. Must be between 1 and 2^24 - 1.
+        let lanes = 4;
+
+        // Desired total size of block matrix, in kibibytes (1 KiB = 1024
+        // bytes). Increasing this forces hashing to use more memory in order to
+        // thwart ASIC-based attacks. Must be >= 8 * lanes.
+        let block_matrix_size = 1 << 16;
+
+        // some extra bytes that can be optionally included
+        let optional_secret_value = [];
```
willemolding, line 56: Should we be using this?

src/holo-crypto/crates/holo-crypto/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+            ARGON2_MIN_SALT_LEN..=ARGON2_MAX_SALT_LEN => Ok(()),
+            _ => Err(HoloCryptoError::Custom(
+                "Salt exceeds maximum allowable number of bytes".to_string(),
+            )),
+        }?;
+
+        // generate a 32 byte password hash using the Argon2 algo
+        let mut secretkey_bytes = [0; SECRET_KEY_LENGTH];
+
+        //The number of block matrix iterations to perform. Increasing
```
willemolding, line 40: Argon2 algorithm parameters taken from Yegor's test implementation so they have not had much thought put in to them by me. A discussion needs to happen around the proper selection of these values (if there hasn't been already)

src/holo-crypto/crates/holo-crypto/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+            &passphrase_bytes,
+            &salt_bytes,
+            &optional_secret_value,
+            ARGON2_ADDITIONAL_DATA,
+        );
+
+        // use this as the secret key to then derive a public key from using Ed25519
+        // this seems kinda strange but is taken out of the examples for the Ed25519 crate
+        // which uses a cryptographic random number instead
+        let secret = SecretKey::from_bytes(&secretkey_bytes)?;
```
willemolding, line 75: Can I get a sanity check here? Then I can remove this comment about it seeming weird

src/holo-crypto/crates/holo-crypto/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+        let public = (&secret).into();
```
willemolding, line 76: This is strange usage of the `From` trait but this is how a public key can be derived from a secret key

src/holo-crypto/crates/holo-crypto/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+                "Salt exceeds maximum allowable number of bytes".to_string(),
+            )),
+        }?;
+
+        // generate a 32 byte password hash using the Argon2 algo
+        let mut secretkey_bytes = [0; SECRET_KEY_LENGTH];
+
+        //The number of block matrix iterations to perform. Increasing
+        // this forces hashing to take longer. Must be between 1 and 2^32 - 1.
+        let passes = 2;
```
yegortimoshenko: TBH, I would inline all of that into `Argon2::new` call. Or at the very least, define as constants.

lukateras (changes requested):
Reimplementing cryptography is the wrong way to go about this. Instead, just depend on `holo-config-core` crate and use admin derivation function defined there.

PR #95, opened 2019-10-23
1897 - As an HP Admin user I can see card-style overviews of Hosting, Earnings, and, HoloFuel (Home)

robbiecarlton:


__mocks__/react-router-dom.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
```
robbiecarlton, line 1: See this file and how I'm using it in `Dashboard.test.js`

Because of this I had to add an `unmock` to a bunch of files that need the real `ReactRouter`, which unfortunately made this PR a little noisy. But the new pattern in `Dashboard.test.js` I think is much better for handling Links, and we should move over to that and ultimately be able to remove the `unmock`s.

src/models/Happ.js
```diff
@@ -0,0 +1,11 @@
+export const defaultHapp = {
```
robbiecarlton, line 1: I'm just using files in the models folder to collect any code that knows specific details about specific graphql types. So this refers to the Happ type in graphql, and similarly there is one for Transaction. We can just expand as needed.

src/pages/Dashboard/Dashboard.test.js
```diff
@@ -0,0 +1,147 @@
...
+    const { getByText } = await renderAndWait(<MockedProvider mocks={mocks} addTypename={false}>
+      <Dashboard />
+    </MockedProvider>, 0)
+
+    expect(getByText('Hi!')).toBeInTheDocument()
+
+    const hosting = getByText('+ Host your first app')
+    expect(hosting).toBeInTheDocument()
+    fireEvent.click(hosting)
+    expect(mockNavigateTo).toHaveBeenCalledWith('/browse-happs')
```
robbiecarlton, line 68: See my comments on `react-router-dom` mock at the top of the PR re this pattern. Here we can see the benefit we get of not having to actually render the router, and being able to easily test that the link does what we expect.

JettTech (approved):
Overall LGTM! Just resolve the package.json andt hen I think it's ready for merge. Also, thanks for taking the time to look into a better router approach in testing.

PR #96, opened 2019-10-23
Rename test:all to test:ci

robbiecarlton:
@zo-el I want to change this back because our CI script expects this script to be called `test:ci`

Will this break anything that you're doing?


zo-el (approved):
:+1:

PR #98, opened 2019-10-28
Wasm crypto with holo config

willemolding:
# holo-crypto WASM

Second pass at a WASM lib that implements some of the crypto algorithms required by Holo.

## Consuming the WASM lib

The library can be installed and imported by other javascript modules in the project by first building the lib by running `wasm-pack build` in the holo-crypto-wasm dir and then adding the following to a package.json
```json
"dependencies": {
    ...
    "holo-crypto-wasm": "file:../holo-crypto-wasm/pkg",
}
``` 
and then imported as normal in js/ts
```javascript
import { AdminKeyPair, WebUserKeyPair } from "holo-crypto-wasm";
```

## API

The lib exposes a very simple object style API to javascript. A KeyPair (either admin or webuser) can be created by instantiating the appropriate object and passing a passphrase, email, salt etc. The function sign() can then be called on this object to sign a string payload. The public and private keys never leave the WASM memory space. 

```javascript
import { AdminKeyPair } from "holo-crypto-wasm";
const kp = new AdminKeyPair(
      "some@email.test",
      "p4ssw0rd",
      "HCA01adefasdtbeodbgahfawnf"
    )
const sig = kp.sign("some message")
```

Both these functions will throw Javascript exceptions with human readable error information if the Rust code errors.

## Crypto Considerations

This crate makes use of the `holo-config-core` crate
## WASM sizes and Benchmarking

The holo-crypto-wasm crate is able to build to a WASM binary that is about 243kB with some optimizations enabled. This is certainly small enough to load client side without much trouble This possibly could be reduced further with aggressive optimization. 

---

Some trivial benchmarks can be obtained by running the included benchmarks page in the browser. 

function | avg time (ms)
---------- | -----------------
Keypair.new() | 324.8
keypair.sign() | 0.28

This suggests both the key generation and signing are probably fast enough for production use


src/holo-crypto-wasm/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+console_error_panic_hook = { version = "0.1.1", optional = true }
+
+# `wee_alloc` is a tiny allocator for wasm that is only ~1K in code size
+# compared to the default allocator's ~10K. It is slower than the default
+# allocator, however.
+#
+# Unfortunately, `wee_alloc` requires nightly Rust when targeting wasm for now.
+wee_alloc = { version = "0.4.2", optional = true }
+
+holo-config-core = { git = "https://github.com/Holo-Host/holo-config", rev = "d70f1f9b8fd7ab76df739abd12b9019ebef3ac44" }
```
willemolding, line 29: This can change to reference develop branch as soon as Holo-Host/holo-config#11 is merged

src/holo-crypto-wasm/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+#[wasm_bindgen]
+pub struct WebUserKeyPair(Keypair);
+
+const WEBUSER_ARGON_ADDITIONAL_DATA: &[u8] = b"holo-crypto webuser ed25519 key v1";
+
+#[wasm_bindgen]
+impl WebUserKeyPair {
+    #[wasm_bindgen(constructor)]
+    pub fn new(
+        _email: &str,
```
willemolding, line 65: Awaiting confirmation as to exactly how these parameters should be combined/passed to the argon2 algorithm

willemolding:
It would be great to have the required build deps (rust + wasm target, wasm-pack) included in the nix thing defined in the root. Also to have some custom commands (e.g. build-wasm-crypto, ..) defined to make it easy for consumers to build it. Unfortunately this is well beyond my nix abilities :(

JettTech:
This work has shifted and this module is no longer most current. I will be closing this PR.

PR #101, opened 2019-10-29
(Do not merge) Testing setup for Joels review

robbiecarlton:


.env.example
```diff
@@ -1,3 +1,3 @@
 REACT_APP_DNA_INTERFACE_URL=ws://localhost:3400
 REACT_APP_HPOS_URL=ws://localhost:3500
-REACT_APP_AGENT_ID=HcScJyCCFZ83Mgt74vxzU5fabdtw8vcmbc7nkkY9HNabig36PUinfDzjebbu54r
\ No newline at end of file
+REACT_APP_CONDUCTOR_CONFIG_PATH=/Users/robbiecarlton/holo/hp-admin5/conductor-config.toml
```
robbiecarlton: this is not used, ignore

src/holochainClient.js
```diff
@@ -1,9 +1,10 @@
 import { connect as hcWebClientConnect } from '@holochain/hc-web-client'
 import { get } from 'lodash/fp'
 import mockCallZome from 'mock-dnas/mockCallZome'
+import { findInstanceForAgent } from 'utils/conductorConfig'
```
robbiecarlton, line 4: @zo-el this is the file that uses `findInstanceForAgent`, see `conductorConfig` as well

src/utils/conductorConfig.js
```diff
@@ -0,0 +1,22 @@
...
+export function getAgent (index = 0) {
+  if (index >= config.agents.length) throw new Error(`There are less than ${index - 1} agents in the config`)
+  return {
+    id: config.agents[index].public_address,
+    nickname: config.agents[index].name,
+    conductorId: config.agents[index].id
+  }
+}
+
+export function findInstanceForAgent (instanceId, agentIndex = 0) {
```
robbiecarlton, line 13: @zo-el - this is the function we need for `holochainClient` that we currently read the config toml to make it work. If you can see a way to do this just using conductor calls, that would be awesome.

PR #106, opened 2019-11-07
3799 - We can build HoloFuel as a separate app and a version of HoloFuel inside of HPAdmin

robbiecarlton:


src/holofuel/HFRouter.test.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
```
robbiecarlton, line 1: I wanted to test all 3 cases for this component because it's the kind of thing that will otherwise get broken in the course of everyday use (we might break one of the modes and not notice for a long time if we're just always using a different mode).

This test is a little annoying, and keeping it up to date (adding new links and assertions as we add new pages to HF) will be a bit of a chore, but I think overall it's worth it.

JettTech (approved):
Overall, this is a good approach and great update! I'd like to talk over the testing implementation together a bit and am curious to see if we can find a few ways to ease the testing overhead, but I say let's give it a go and talk along the way.  This PR looks good. :tada:

PR #107, opened 2019-11-08
0531 test refresh

JettTech:


src/pages/MyProfile/MyProfile.js
```diff
@@ -61,7 +77,9 @@ const MyProfile = ({ history: { push } }) => {
         </small>}
       </label>
 
-      <label styleName='field'>
+      {/* TODO: Disuss email display options with C & Z. - Do we want to display the email here at all ? */}
```
robbiecarlton: is this captured in pivotal?

src/pages/MyProfile/MyProfile.js
```diff
@@ -72,9 +90,10 @@ const MyProfile = ({ history: { push } }) => {
         {errors.email && <small styleName='field-error'>
           You need to provide a valid email address.
         </small>}
-      </label>
+      </label> */}
 
-      <label styleName='field'>
+      {/* NB: This field is not optional in the first release. A new pw would create new a pubkey / agent... */}
```
robbiecarlton: let's just remove all this commented code.

src/pages/MyProfile/MyProfile.test.js
```diff
@@ -39,19 +43,19 @@ describe('Rendering', () => {
     expect(input).toBeInTheDocument()
   })
 
-  it('should render the email input', async () => {
-    const { getByLabelText } = renderMyProfile()
-    const input = getByLabelText('Email')
+  // it('should render the email input', async () => {
```
robbiecarlton: use it.skip rather than commenting out tests.

It gives us more visibility in the test report.

src/pages/MyProfile/MyProfile.test.js
```diff
@@ -80,55 +84,55 @@ describe('Validation', () => {
...
-
-    await act(async () => {
-      fireEvent.click(getByText('Save Changes'))
-      await wait(0)
-    })
-
-    const error = queryByText('You need to provide a valid email address.')
-    expect(error).not.toBeInTheDocument()
-  })
+  // it('should reject empty email', async () => {
```
robbiecarlton: it.skip

src/pages/MyProfile/MyProfile.test.js
```diff
@@ -150,9 +154,14 @@ describe('Validation', () => {
...
     const nameImput = getByLabelText('Name')
-    fireEvent.change(nameImput, { target: { value: 'Alice' } })
-    const emailInput = getByLabelText('Email')
-    fireEvent.change(emailInput, { target: { value: 'alice@example.com' } })
+    await act(async () => {
+      fireEvent.change(nameImput, { target: { value: 'Alice' } })
+      await wait(0)
+    })
+
+    // NOTE: Currently we will not be updating email.
```
robbiecarlton: let's just remove the commented code.

src/utils/test-utils.js
```diff
@@ -9,3 +11,28 @@ export async function renderAndWait (ui, ms = 0, options = {}) {
   })
   return queries
 }
+
+export const renderAndWaitWithRouter = async (ui, ms = 0, options = {}) => {
```
robbiecarlton: so my preference here would be that we just combine them in the code, and export them separately. There's something about this being one monolithic function that is unwieldy. Like, how do we then combine it with the modal code that you have in the other pr? We end up with `renderAndWaitWithRouterAndModal` and etc etc.

Instead if we just export `wrapInRouter` and `renderAndWait` we can compose them in the test

`renderAndWait(wrapInRouter(<Component />))`

robbiecarlton (changes requested):

src/pages/MyProfile/MyProfile.js
JettTech: So, I brought it up in stand-up yesterday, but currently the decision is to not focus on the profile as it won't be included in the first iteration.  I will check and make a sure there is a backlog item to earmark this discussion though.

src/pages/MyProfile/MyProfile.test.js
JettTech: I was leaving this code visible, but commented out, as we likely will be needing it in the near future (after our first release)... but I also like the cleanliness without the noise of the comments. We can remove it for now. :+1:

src/pages/MyProfile/MyProfile.test.js
JettTech: The commented out test code relates to the commented out component code.  Since we're deleting the component commented code, I'll go ahead and just delete this code respectively, instead of skipping it, as this would be the most accurate overview of the tests.

robbiecarlton (approved):

PR #108, opened 2019-11-11
Updated the nix process

zo-el:


default.nix
```diff
@@ -89,28 +89,75 @@ in
...
+
+    installPhase = ''
+      mv hp-admin $out
+    '';
+
+    checkPhase = ''
+      npm run test:ci
+    '';
+
+    doCheck = false;
```
samrose, line 114:
```suggestion
    doCheck = false;
    meta.platforms = stdenv.lib.platforms.linux;
```

default.nix
```diff
@@ -89,28 +89,75 @@ in
...
+      mv holofuel $out
     '';
 
     checkPhase = ''
-      npm run test:all
+      npm run test:ci
     '';
 
-    doCheck = true;
+    doCheck = false;
```
samrose, line 161:
```suggestion
    doCheck = false;
    meta.platforms = stdenv.lib.platforms.linux;
```

samrose (changes requested):

default.nix
```diff
@@ -89,28 +89,77 @@ in
     buildInputs = [
       run-unit-test
       build-hp-admin
```
peeech, line 91: I am not sure what is the purpose of this import here. `build-hp-admin` is essentially a shell script that installs npm modules and runs the `build` command. Same functions are encoded in `preConfigure` and `buildPhase` below. Would love to hear an opinion from @samrose or @yegortimoshenko on this.

default.nix
```diff
@@ -89,28 +89,77 @@ in
...
+      libjpeg
+      libpng
+      libuuid
+      pango
+      pixman
+    ];
+
+    buildInputs = [
+      run-unit-test
+      build-holofuel
```
peeech, line 138: Same comment here re: is this build phase necessary

default.nix
zo-el: @peeech This was just added for the users of the repo(hAppy team) to build their project using nix.

peeech: ok, thanks for clarification

default.nix
peeech: resolved

pjkundert:
Looks good to me.  If we want to use it against current envoy, we'd need to add an artifact from hclient (as per: https://github.com/Holo-Host/holo-nixpkgs/blob/201909/demo-cherry-pick/overlays/holo-nixpkgs/holofuel-app/default.nix), but we're working toward supporting the *new* "chaperone" SDK, so this is probably not necessary.

It looks good!

pjkundert (approved):
This is sufficient to get the `holofule-app` UI into the flow; we'll need to update it to a more recent branch, later, as this one doesn't work (even for the old hClient / envoy), due to needing an updated hClient.  That's OK for now, though.

PR #110, opened 2019-11-11
Fixes to the testing issues

zo-el:
Fixed Issues:
- `cp` errors 
- `holochain instances not active`
- `holochain public token issues`
- Tests not running individually  

Bug Discovered
- DNA returns values but it is not loading in the tests?


src/utils/integration-testing/runConductorWithFixtures.js
```diff
@@ -41,15 +42,21 @@ export default function runConductorWithFixtures (testFn) {
...
+                    // if (stderr) {
+                    //   console.error(e)
+                    //   throw new Error('Error coping Snapshot Storage dir into Default Storage dir: ')
+                    // } else {
+                    //   console.log('Copied Snapshot Storage into Default Storage!')
+                    //   storageDir = 'Snapshot Storage Direcotry'
+                    //   resolve(storageDir)
+                    // }
+                    await exec(`rm -rf ${process.env.REACT_APP_DEFAULT_STORAGE} && mkdir ${process.env.REACT_APP_DEFAULT_STORAGE}`)
+                    await ncp(process.env.REACT_APP_DEFAULT_STORAGE, process.env.REACT_APP_STORAGE_SNAPSHOT, (e)=>{
```
JettTech, line 55: Good idea... but how does this differ technically from the recursive copying that first removes from the destination (ie: 'cp -r --remove destination' )?  Also, ....just curious, what is the motive for using the ncp module instead of the bash script?

scripts/wait-for-conductor.js
```diff
@@ -31,7 +32,10 @@ async function waitForConductor (interval = 10000) {
   while (!isUp) {
     try {
       console.log('Checking if conductor is up')
-      const result = await axios.post('http://localhost:3300/admin/agent/list', {})
+      // We need to check if the instance is live
+      // By calling an admin call does not mean that the instances are active
+      const agent1 = getAgent(agent1Index)
+      const result = await axios.post(`http://localhost:3300/${agent1}::<happ_id>-holofuel/transactions/whoami`, {})
```
JettTech, line 38: Good catch! :tada:

config.nix
```diff
@@ -7,13 +7,13 @@
...
     keystore_file = "agent.key";
-    public_address = "HcScJyCCFZ83Mgt74vxzU5fabdtw8vcmbc7nkkY9HNabig36PUinfDzjebbu54r";
+    public_address = "HcScIio76M8cbDqhrsZKp37W8YZ4y8X9v9epzbyhRzqpjeb5sigdC4puN6b5koi";
   }
   {
     id = "hp-admin-agent-2";
     name = "HP Admin 2";
     keystore_file = "agent-2.key";
-    public_address = "HcScJNpnC8kabp59w89Z9GhfHw7VdycrmZP7yaBAP7i3sxynx6KO8Pvm3ai6iiz";
+    public_address = "HcScjSPMA9g46x6nosEih9thqwvesx53udZEkqMfqcwnbw9sZKPSyIFNvtfxrqi";
```
JettTech, line 16: ...But of course we need v.0.0.32 agent keys! :100: :pray:

JettTech (approved):
Great finds and fixes.  Also, I left a comment about the remove and copy approach that I'd like to discuss a bit further. Overall, very helpful updates!!

PR #111, opened 2019-11-11
2041 - As a HoloFuel user viewing my Transaction History, I can toggle between views

robbiecarlton:


src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -58,58 +62,113 @@ function useFetchCounterparties () {
...
+
+  const isToday = momentDate => momentDate.isSame(today, 'd')
+  const isYesterday = momentDate => momentDate.isSame(yesterday, 'd')
+
+  const getLabel = ({ timestamp }) => {
+    const momentDate = moment(timestamp)
+    if (isToday(momentDate)) return 'Today'
+    if (isYesterday(momentDate)) return 'Yesterday'
+    return momentDate.format('MMMM Do')
+  }
```
JettTech: When considering the timestamp label approach for a similar need in Inbox, I went ahead and wrote this logic into the holofuel interface and returned an additional entry for this label in the transaction gql type, in order to avoid nesting that data into a page or component.  This makes it much easier for all pages to readily have all the data they need to sort through the transactions as done previously.  Take a look and let's discuss whether we'd like to reference that data here too.

JettTech (changes requested):
Let's come to a decision about the way we'd like to implement the timestamp labels as well as how much logic we'd like to extract vs leave specialized and/or within the component. Also, I am receiving one failing unit test.  

Overall, though, this looks great. Nice test coverage on this one. :+1:

PR #113, opened 2019-11-13
6628 - As a HoloFuel User, I can view my Inbox Page (v0.1.2)

JettTech:


package.json
```diff
@@ -7,7 +7,7 @@
     "start:holofuel": "REACT_APP_HOLOFUEL_APP=true npm start",
     "build": "node scripts/build.js",
     "build:holofuel": "REACT_APP_HOLOFUEL_APP=true node scripts/build.js",
-    "hc:start": "mkdir -p tmp && holochain -c ./conductor-config.toml",
+    "hc:start": "mkdir -p tmp && holochain -c ./conductor-config.toml &",
```
robbiecarlton, line 10: just wanted to check that this was an intentional change and is meant for this PR.

src/components/icons/AddIcon.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
```
robbiecarlton, line 1: Are these the final icons that are coming from Zoes designs? As in: The svg is being exported from her design files?

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -19,33 +19,37 @@ export async function getTxCounterparties (transactionList) {
...
     amount: amount || event.Request.amount,
     counterparty: {
       id: counterpartyId || event.Request.from
     },
     direction: stateDirection,
-    status: STATUS.pending,
+    status: status || STATUS.pending,
     type: TYPE.request,
     timestamp: eventTimestamp,
+    dateLabel: date,
```
robbiecarlton: remove the dateLabel field everywhere in this file

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -19,33 +19,37 @@ export async function getTxCounterparties (transactionList) {
...
+  const { date } = formatDateTime(eventTimestamp)
   return {
     id: origin,
     amount: amount || event.Request.amount,
     counterparty: {
       id: counterpartyId || event.Request.from
     },
     direction: stateDirection,
-    status: STATUS.pending,
+    status: status || STATUS.pending,
```
robbiecarlton, line 30: is `status` ever going to be something other than `'pending'` here? If not, we should leave it as is for clarity. Wondering how this worked before if not.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -137,8 +145,12 @@ function presentTransaction (transaction) {
       throw new Error('Completed event did not have a Receipt or Cheque event')
     }
     case 'rejected': {
-      // We have decided not to return the reject case into the Ledger
-      break
+      // We have decided to show this **only** in the inbox page via the recent transactions filter
+      console.log(' >>>>>>>> INSIDE THE REJECTED presentTransaction Case <<<<<< EVENT : ', event)
```

src/graphql-server/schema.graphql
```diff
@@ -62,6 +62,7 @@ type Transaction {
   status: TransactionStatus
   type: TransactionType
   timestamp: String
+  dateLabel: String
```
robbiecarlton: remove `datelabel` here an in the other graphql files

src/graphql-server/schema.graphql
```diff
@@ -112,6 +113,7 @@ type Query {
   hostPricing: HostPricing
   holofuelCompletedTransactions: [Transaction]
   holofuelWaitingTransactions: [Transaction]
+  holofuelRecentTransactions: [Transaction]
```
robbiecarlton: Given it would return a transaction from eg 7 years ago, is this the best name for this endpoint? I get that we call these 'recent' transactions on the page, but really it's just 'nonPending' transactions right? There's no time limit on the results here.

src/holochainClient.js
```diff
@@ -3,7 +3,7 @@ import { get } from 'lodash/fp'
 import mockCallZome from 'mock-dnas/mockCallZome'
 
 // This can be written as a boolean expression then it's even less readable
-const developmentMockDnaConnection = true // this is the value MOCK_DNA_CONNECTION will have in the dev server
+const developmentMockDnaConnection = false // this is the value MOCK_DNA_CONNECTION will have in the dev server
```
robbiecarlton: keep this true on develop

src/holofuel/components/Button/Button.module.css
```diff
@@ -33,3 +33,31 @@
...
+  color: rgba(44, 64, 90, 0.5);
+  font-family: 'CircularStd', sans-serif;
+  font-size: 16px;
+  line-height: 22px;
+  letter-spacing: -0.4px;
+  border-radius: 3px;
+}
+
+.toggle-selected {
+  /* composes: toggle; */
```
robbiecarlton, line 50: remove comment, or better yet, uncomment it and remove the redundancy :D

src/holofuel/components/Header/Header.module.css
```diff
@@ -11,8 +15,10 @@
   border: none;
   display: flex;
   align-content: center;
-  background-color: rgb(200,200,200);
-  color: var(--color-gray);
+  background-color: #FFFFFF;
+  color: #000000;
+  /* background-color: rgb(200,200,200); */
```
robbiecarlton: remove comments

src/holofuel/components/Header/Header.module.css
```diff
@@ -32,20 +38,26 @@
   z-index: 99;
   border: none;
   padding: 5px;
-  background-color: var(--color-tundora);
+  background-color: #FFFFFF;
+  /* background-color: var(--color-tundora); */
```
robbiecarlton: remove comment

src/holofuel/components/Header/Header.module.css
```diff
@@ -32,20 +38,26 @@
...
 }
 
 .menu-icon {
   width: 22px;
   height: 22px;
-  background-color: var(--color-tundora);
+  background-color: #FFFFFF;
+  /* background-color: var(--color-tundora); */
```
robbiecarlton: remove comment

src/holofuel/components/Header/Header.module.css
```diff
@@ -11,8 +15,10 @@
   border: none;
   display: flex;
   align-content: center;
-  background-color: rgb(200,200,200);
-  color: var(--color-gray);
+  background-color: #FFFFFF;
```
robbiecarlton: let's just use `background-color: white;` for these. More readable.

src/holofuel/components/Header/Header.module.css
```diff
@@ -32,20 +38,26 @@
...
-  display: flex;
-  background-color: var(--color-tundora);
+  height: 50px;
   padding: 5px 10px;
   align-items: center;
+  display: flex;
+  background-color: #FFFFFF;
+  box-shadow: 0px 2.5px 10px rgba(0, 0, 0, 0.1);
+  backdrop-filter: blur(27.1828px);
+  /* background-color: var(--color-tundora); */
```
robbiecarlton: remove comment

src/holofuel/components/Header/Header.module.css
```diff
@@ -54,7 +66,8 @@
   flex-grow: 1;
   line-height: 28px;
   font-size: 16px;
-  color: rgb(200,200,200);
+  color: #000000
```
robbiecarlton: `color: black;`

src/holofuel/components/Header/Header.module.css
```diff
@@ -54,7 +66,8 @@
...
+  /* color: rgb(200,200,200); */
```
robbiecarlton: remove comment

src/holofuel/components/Header/Header.module.css
```diff
@@ -73,17 +86,17 @@
...
   padding: 10px;
   border: none;
+  background-color: #FFFFFF; /* rgb(140,140,140); */
+  /* background-image: linear-gradient(rgb(100,100,100), rgb(255,255,255)); */
 }
 
-.sub-title {
-  font-size: 20px;
+.page-title {
+  /* font-size: 20px; */
```
robbiecarlton: remove comment

src/holofuel/components/Header/Header.module.css
```diff
@@ -73,17 +86,17 @@
...
-  background-color: rgb(140,140,140);
-  /* background-image: linear-gradient(rgb(100,100,100), rgb(255,255,255)); */
+.page-header {
   text-align: center;
-  color: white;
+  color: #000000;
   padding: 10px;
   border: none;
+  background-color: #FFFFFF; /* rgb(140,140,140); */
+  /* background-image: linear-gradient(rgb(100,100,100), rgb(255,255,255)); */
```
robbiecarlton: remove comment

src/holofuel/components/Header/Header.module.css
```diff
@@ -73,17 +86,17 @@
...
-.sub-header {
-  background-color: rgb(140,140,140);
-  /* background-image: linear-gradient(rgb(100,100,100), rgb(255,255,255)); */
+.page-header {
   text-align: center;
-  color: white;
+  color: #000000;
   padding: 10px;
   border: none;
+  background-color: #FFFFFF; /* rgb(140,140,140); */
```
robbiecarlton: `color: white;` and remove comment

src/holofuel/components/Modal/Modal.module.css
```diff
@@ -1,13 +1,16 @@
...
+  max-width: 90%;
   margin: 0 auto;
   top: 30%;
   background-color: #fff;
   border: 1px solid #000;
-  border-radius: 8px;
-  box-shadow: 4px 4px 8px rgba(0, 0, 0, .5);
-  padding: 32px;
+  border-radius: 1px;
+  /* box-shadow: 4px 4px 8px rgba(0, 0, 0, .5); */
```
robbiecarlton: remove comment

src/holofuel/components/Jumbotron/Jumbotron.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import cx from 'classnames'
+// not importing styles here breaks tests. This is a hack.
```
robbiecarlton: is that true? I think we can just do an `import './Jumbotron.module.css'`

src/holofuel/components/NullStateMessage/NullStateMessage.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import cx from 'classnames'
+// not importing styles here breaks tests. This is a hack.
```
robbiecarlton: same question

src/holofuel/components/Jumbotron/Jumbotron.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
```
robbiecarlton, line 1: Should `Jumbotron` and `NullStateMessage` be separate components with their own tests and stylesheets etc?

I think more likely they should be local components within the Inbox page file.

src/holofuel/components/PageDivider/PageDivider.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
```
robbiecarlton, line 1: Probably doesn't need to be in its own file. And probably doesn't actually need to be a component. This is just an H6

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.module.css
```diff
@@ -4,6 +4,7 @@
   word-break: break-word;
   height: 100%;
   box-sizing: border-box;
+  background-color: #ECECEC;
```
robbiecarlton: if this is a final color we're using, can you put it in the `colors.css` file and import it as a variable?

zo-el (approved):
LGTM :+1:

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -58,31 +68,122 @@ function useFetchCounterparties () {
...
-    setModalTransaction(modalTransaction)
+    if (!isEmpty(transaction) && action !== '') return setModalTransaction(modalTransaction)
+    else return setToggleModal(true)
   }
 
-  return <PrimaryLayout headerProps={{ title: pageTitle }}>
+  const [actionsVisible, setActionsVisible] = useState(null)
+  const actionsClickWithTx = transaction => setActionsVisible(transaction)
+
+  const toggleButtons = [{ view: 'pending', label: 'Pending' }, { view: 'recent', label: 'Recent' }]
```
robbiecarlton: if you're going to use the VIEW constants, you should use them here as well `view: VIEW.pending` etc

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -58,31 +68,122 @@ function useFetchCounterparties () {
...
+  let displayTransactions = []
+  switch (inboxView) {
+    case VIEW.pending:
+      displayTransactions = actionableTransactions
+      break
+    case VIEW.recent:
+      displayTransactions = recentTransactions
+      break
+    default:
+      displayTransactions = actionableTransactions
```
robbiecarlton, line 125: I'd throw an error here. If we hit this clause, something has gone wrong.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -58,31 +68,122 @@ function useFetchCounterparties () {
...
+      break
+  }
 
-    {!isTransactionsEmpty && <div styleName='transaction-list'>
-      {transactions.map(transaction => <TransactionRow
-        transaction={transaction}
-        showConfirmationModal={showConfirmationModal}
-        key={transaction.id} />)}
+  const isPendingTransactionsEmpty = isEmpty(actionableTransactions)
```
robbiecarlton: We should pick one thing to call these transactions in the code. I could go either way. We call them 'pending' in the UI, but we already called them 'actionable' and that would be a bunch of changes, plus I'm not convinced we'll keep the name 'pending' in the UI. It's nice when the name for things is the same in the code as in the UI, but especially at the beginning like this, it's not always practical. So I think I've talked myself into calling them 'actionable' everywhere for now.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -58,31 +68,122 @@ function useFetchCounterparties () {
...
+  const isDisplayTransactionsEmpty = isEmpty(displayTransactions)
+  const pageTitle = `Inbox${isPendingTransactionsEmpty ? '' : ` (${actionableTransactions.length})`}`
```
robbiecarlton: eg: this line is noticeably harder to read because of the different names for the same thing.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -58,31 +68,122 @@ function useFetchCounterparties () {
...
+
+  // Sort the transactions by (semantic) date
+  const transactionsByDate = groupBy('dateLabel', displayTransactions)
+  const transactionsByDateKeys = Object.keys(transactionsByDate)
```
robbiecarlton: replace this with `partitionByDate` from `utils`

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -58,31 +68,122 @@ function useFetchCounterparties () {
...
+
+  return <PrimaryLayout headerProps={{ title: pageTitle }} inboxCount={actionableTransactions.length}>
+    <Jumbotron
+      className='inbox-header'
+      title={`${presentHolofuelAmount(holofuelBalance)} HF`}
+      titleSuperscript='Balance'isTransactionsEmpty
+    >
+      <Button styleName='new-transaction-button' onClick={() => showConfirmationModal()}>
+        {/* TODO: Resolve issue with path for the ADD icon >> not displayed properly */}
```
robbiecarlton: I think we should just remove the TODO. This is a task for the UI implementation ticket and we will naturally catch it then.

(i don't know if you've noticed, but I'm not a big fan of TODO comments ;) )

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -58,31 +68,122 @@ function useFetchCounterparties () {
...
+  const showConfirmationModal = (transaction = {}, action = '') => {
     const modalTransaction = { ...transaction, action }
-    setModalTransaction(modalTransaction)
+    if (!isEmpty(transaction) && action !== '') return setModalTransaction(modalTransaction)
+    else return setToggleModal(true)
   }
 
-  return <PrimaryLayout headerProps={{ title: pageTitle }}>
+  const [actionsVisible, setActionsVisible] = useState(null)
+  const actionsClickWithTx = transaction => setActionsVisible(transaction)
```
robbiecarlton: this is redundant
`actionsClickWithTx` is the same function as `setActionsVisible`

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -58,31 +68,122 @@ function useFetchCounterparties () {
...
+      </NullStateMessage>
+    </>}
+
+    {!isDisplayTransactionsEmpty && !isEmpty(transactionsByDateKeys) && <div className='transaction-by-date-list'>
+      {transactionsByDateKeys.map(transactionDate => <div key={transactionDate}>
+        <PageDivider title={transactionDate} />
+        <div styleName='transaction-list'>
+          {transactionsByDate[transactionDate].map(transaction => <TransactionRow
+            transaction={transaction}
+            actionsVisible={actionsVisible}
```
robbiecarlton: couple of things. 
First we should be storing just the transaction id in actionsVisible, rather than the whole transaction. It's just safer and cleaner to compare ids than to compare references to objects.

Second I would do the comparison here

`showActionButtons={actionsVisible === transaction.id}`

and then TransactionRow doesn't need to know how the parent component manages which transactions are showing action buttons. This makes the code easier to maintain, refactor, and make feature changes..

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -91,46 +192,90 @@ export default function Inbox () {
...
+
+  const actionsClick = () => actionsClickWithTx(transaction)
+  const handleCloseReveal = () => {
+    if (!isEmpty(actionsVisible) && actionsVisible === transaction) return actionsClickWithTx(null)
+    else if (!isEmpty(actionsVisible) && actionsVisible !== transaction) return actionsClickWithTx(transaction)
+    else return actionsClickWithTx(null)
+  }
+
+  let agent
+  if (counterparty.id === whoami.id) agent = whoami
```
robbiecarlton: use curly braces for non returning `ifs` as per the style guide

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -91,46 +192,90 @@ export default function Inbox () {
...
+  else agent = counterparty
 
   const isOffer = type === TYPE.offer
   const isRequest = !isOffer
 
-  const { date, time } = presentDateAndTime(timestamp)
-
-  const story = isOffer ? ' is offering' : ' is requesting'
+  let story
+  if (inboxView === VIEW.pending) story = isOffer ? ' is offering' : ' is requesting'
```
robbiecarlton: rather than having the TransactionRow care about the state of the inbox, it makes more sense to me that the TransactionRow just knows whether this is a pending transaction or not. I think we can even get that from the transaction object, so no need to pass in inboxView.

I can also see an argument for wanting the parent to be in control of whether this displays as pending, rather than the TransactionRow, in which case you can just pass in

`isPending={inboxView === View.pending}`
or
`isPending={isPending(transaction)}`      // this test could also be done in TransactionRow, I feel mixed about that.
or whatever other condition, as a prop to `TransactionRow`

This is the same principle that motivated my request around actionsVisible: right now `TransactionRow` knows more than it needs to about the inner workings of the Inbox component, which makes it harder to refactor/maintain/change.

src/holofuel/pages/Inbox/Inbox.test.js
```diff
@@ -8,12 +8,13 @@ import apolloClient from 'apolloClient'
 import Inbox, { TransactionRow } from './Inbox'
 import { pendingList } from 'mock-dnas/holofuel'
 import { TYPE } from 'models/Transaction'
-import { presentHolofuelAmount } from 'utils'
+import { presentHolofuelAmount } from 'utils' // , formatDateTime
```
robbiecarlton: remove comment

src/holofuel/pages/Inbox/Inbox.test.js
```diff
@@ -8,12 +8,13 @@ import apolloClient from 'apolloClient'
...
 import { renderAndWait } from 'utils/test-utils'
+// import { title as forwardIconTitle } from 'components/icons/ForwardIcon'
```
robbiecarlton: remove comment

src/holofuel/pages/Inbox/Inbox.test.js
```diff
@@ -167,20 +176,34 @@ describe('TransactionRow', () => {
...
+      const { getByText, getByTestId } = await renderAndWait(<MockedProvider mocks={mocks} addTypename={false}>
         <TransactionRow {...props} />
       </MockedProvider>, 0)
 
+      expect(getByTestId('forward-icon')).toBeInTheDocument()
+      await act(async () => {
+        fireEvent.click(getByTestId('forward-icon'))
+        await wait(0)
+      })
+      // expect(getByText(forwardIconTitle)).toBeInTheDocument()
```
robbiecarlton: remove comment

src/holofuel/pages/Inbox/Inbox.test.js
```diff
@@ -200,4 +223,22 @@ describe('TransactionRow', () => {
       expect(acceptOfferMock.newData).toHaveBeenCalled()
     })
   })
+
+  // describe('Semantic Timestamp Label', () => {
```
robbiecarlton: remove comments

src/holofuel/pages/Inbox/Inbox.test.js
```diff
@@ -200,4 +223,22 @@ describe('TransactionRow', () => {
...
+  //     const { getByText } = await renderAndWait(<MockedProvider mocks={mocks} addTypename={false}>
+  //       <PageDivider title={request.timestamp} />
+  //     </MockedProvider>, 0)
+
+  //     expect(getByText(formatDateTime(request.timestamp.format('MMM D YYYY')))).toBeInTheDocument()
+  //   })
+  // })
 })
+
+// Add'l tests to add & review :
```
robbiecarlton: let's add these tests

robbiecarlton (changes requested):

robbiecarlton:
Let me know if you want to jump on Zoom and discuss any of this

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Yes, it will now sometimes be rejected/declined.

src/graphql-server/schema.graphql
JettTech: True, and I also prefer the separation of naming logic.  I'll update acccordingly.

src/holofuel/components/Jumbotron/Jumbotron.js
JettTech: They were designed such that they could be shared across all the pages in HF that include those styles, including Transactions History, Request/Offer HF, and Home.  With this in mind, I think they should be extracted.

src/holofuel/components/PageDivider/PageDivider.js
JettTech: As it will be referenced and used across at least three separate pages (ie: Inbox, Transaction History, Home), I think we should turn this into a design-system component.  It is simple, but there is not reason to repeat the code in each of these style-sheets.  (IMO, there are a few other components, like the nav badge, that would benefit from this too.  ....I also recognize that this is a design system question and we can talk it about it at more length when we take on that discussion.)

src/holofuel/pages/Inbox/Inbox.test.js
JettTech: Done. :100: :rocket:

src/components/icons/AddIcon.js
JettTech: No, these are placeholder icons.  I am waiting until after our design system talk / sync to begin implementation of the actual icons. Good question though.

robbiecarlton: It will be easy to get this mixed up now that I'm about to merge some work with real icons. Can you rename these components to be `PlaceholderAddIcon` etc?
I know that's annoying. It's just, I just made a component called `PlusIcon` and when this is merged, and you're looking for the icon to use it's not going to be obvious which one is correct. Then when someone is implementing final designs on this page, they can delete the whole `PlaceholderAddIcon` component.

src/holofuel/pages/Inbox/Inbox.js
robbiecarlton: I still think that `actionsVisible` shouldn't be passed into `TransactionRow`, that instead you should pass in a `showActionButtons` prop so that `TransactionRow` doesn't have to know how its parent handles which transactions are shown.

robbiecarlton: Also, sorry I should have thought of this in my first review: `actionsVisible` should be renamed `actionsVisibleId` to be clear that it's an id.

robbiecarlton (changes requested):
Good changes! Just a couple of small things and then it's good to merge.

src/components/icons/AddIcon.js
JettTech: Good point.  I totally agree that it's better to be clear now and avoid any confusion leading to code debt later. :100:

JettTech:
@robbiecarlton - Per our mm message, I'm going to merge this now.  FYI: Renamed the actionsVisible to actionsVisibleId as requested. Good point regarding the name clarity.  :+1:

PR #114, opened 2019-11-14
1960 - As a HoloFuel user on the Send and Request screens, I can see my most recent transaction partners

robbiecarlton:


src/holofuel/components/RecentCounterparties/RecentCounterparties.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
```
robbiecarlton, line 1: Once we merge `CreateRequest` and `CreateOffer` into a single page, this component should be in that file, not it's own component.

src/holofuel/components/RecentCounterparties/RecentCounterparties.js
```diff
@@ -0,0 +1,33 @@
...
+import { isEmpty } from 'lodash/fp'
+import cx from 'classnames'
+import HashIcon from 'holofuel/components/HashIcon'
+import { presentAgentId } from 'utils'
+import './RecentCounterparties.module.css'
+
+export default function RecentCounterparties ({ agents, selectedAgentId, selectAgent, className }) {
+  if (isEmpty(agents)) return null
```
robbiecarlton, line 9: I've never figured out how I want to handle this. Should the component be responsible for deciding whether to display itself, or should this logic be in the parent?

I go back and forth, and it might be that there isn't a single answer and it varies depending on the situation.

Curious your thoughts on this.

src/holofuel/pages/CreateRequest/CreateRequest.test.js
```diff
@@ -227,6 +229,57 @@ describe('CreateRequest', () => {
     expect(within(getByTestId('counterparty-nickname')).getByText('Loading')).toBeInTheDocument()
   })
 
+  it('renders a clickable list of recent counterparties', async () => {
```
robbiecarlton, line 232: obv the duplication between this and `CreateOffer.test` is pretty egregious, but I think this is the best solution til we merge them. Definitely feeling ready to merge them :D

src/holofuel/pages/CreateRequest/CreateRequest.test.js
```diff
@@ -227,6 +229,57 @@ describe('CreateRequest', () => {
...
+    const { getAllByTestId, getByLabelText } = await renderAndWait(<MockedProvider mocks={mocks} addTypename={false}>
+      <CreateRequest history={{}} />
+    </MockedProvider>)
+
+    const agentRows = getAllByTestId('agent-row')
+
+    const agent1Row = agentRows.find(agentRow => {
+      const { queryByText } = within(agentRow)
+      return !!queryByText(agent1.nickname)
+    })
```
robbiecarlton, line 263: I don't love this way of getting at agent1Row, but as far as I can tell, it's the best way of doing this.

src/holofuel/components/RecentCounterparties/RecentCounterparties.js
jmday: Sensible default logic in component with parent access to override methods?

JettTech: I agree with @jmday.  The component itself should be have a default state that it can expect, but should also have a override option that it can manage. I also like the thought that this feature being default visible whenever no agents are available, just to notify the user of the fact that they'll only ever need to copy the each counterparty-agent's hash once.

...Maybe we could do something simple like the following, wherein we just set the `displayOn` to a global var in the parent component.  (Maybe this var could eventually be controlled by a settings option, or the like, too.)
```
export default function RecentCounterparties ({ agents, selectedAgentId, selectAgent, className, displayOn }) {
  if (displayOn && isEmpty(agents)) agents = null
  else if (isEmpty(agents)) return null

  return <div styleName='recent-counterparties' className={className}>
    <div styleName='header'>Recent Friends</div>
   {isEmpty(agents) && <div styleName='agent-row'>No Recent Friends</div>}
    {agents && agents.map(agent => <AgentRow
      agent={agent}
      selected={agent.id === selectedAgentId}
      selectThisAgent={() => selectAgent(agent.id)}
      key={agent.id}
    />)}
  </div>
}
```

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -34,19 +36,28 @@ function useOfferMutation () {
 }
 
 export default function CreateOffer ({ history: { push } }) {
+  const { data: { holofuelHistoryCounterparties: agents } = {} } = useQuery(HolofuelHistoryCounterpartiesQuery)
+  const recentAgents = agents ? agents.slice(0, 6) : []
```
JettTech: Can we pass the full array of recentAgents into the `RecentCounterparties` and have this display logic live there instead?  As this is display info specific to the child component, I think that it makes sense to live there.  (Also, within a later iterations, we might wish to have a infinite scroll or pagination logic applied to the recent counterparty display and we would need to have access to the full array at that time.)

src/holofuel/pages/CreateRequest/CreateRequest.js
```diff
@@ -33,15 +35,24 @@ function useRequestMutation () {
 }
 
 export default function CreateRequest ({ history: { push } }) {
+  const { data: { holofuelHistoryCounterparties: agents } = {} } = useQuery(HolofuelHistoryCounterpartiesQuery)
+  const recentAgents = agents ? agents.slice(0, 6) : []
```
JettTech: As with CreateOffer, can we pass the full agent array into `RecentCounterparties`, and move line 39 to live there instead?

src/holofuel/pages/CreateRequest/CreateRequest.test.js
JettTech: Totally agree!  Cannot wait to merge them into a single page! :raised_hands:

src/holofuel/pages/CreateOffer/CreateOffer.test.js
```diff
@@ -235,6 +237,57 @@ describe('CreateOffer', () => {
...
+    const mocks = [
+      {
+        request: {
+          query: HolofuelCounterpartyQuery,
+          variables: { agentId: agent1.id }
+        },
+        result: {
+          data: { holofuelCounterparty: agent1 }
+        }
+      },
```
JettTech: Do we need the HolofuelCounterpartyQuery mock here?  If not, can we remove it?

src/holofuel/pages/CreateRequest/CreateRequest.test.js
```diff
@@ -227,6 +229,57 @@ describe('CreateRequest', () => {
...
+    const mocks = [
+      {
+        request: {
+          query: HolofuelCounterpartyQuery,
+          variables: { agentId: agent1.id }
+        },
+        result: {
+          data: { holofuelCounterparty: agent1 }
+        }
+      },
```
JettTech: Same as in CreateOffer test.. Do we need this mock for this test?

src/holofuel/pages/CreateRequest/CreateRequest.test.js
JettTech: Since we know we'll be testing the first agent, we can also expect (and test that) the first row in agentRows is the parent row of agent1 nickname.  If we go with this approach we could also do something like the following to get the full agent1Row:

```
   const { getAllByTestId, getByLabelText } = await renderAndWait(<MockedProvider mocks={mocks} addTypename={false}>
      <CreateOffer history={{}} />
    </MockedProvider>)

    const agentRows = getAllByTestId('agent-row')
    const agent1Row = agentRows[0]
    const { queryByText, getByText } = within(agent1Row)

    expect(queryByText(agent1.nickname)).toBeInTheDocument()
    expect(agent1Row).toBeInTheDocument()

    fireEvent.click(agent1Row)

    expect(getByLabelText('To').value).toEqual(agent1.id)
    expect(getByText('Selected')).toBeInTheDocument()
```

I'm fine with either approach though.

JettTech (changes requested):
Overall, looks great.  Just requested a couple changes, which could lead to a couple conversations. :)  Let me know if you have any Qs.

src/holofuel/components/RecentCounterparties/RecentCounterparties.js
robbiecarlton: you guys are talking about something more complicated than I meant.

I'm just wondering about having this code in the parent component

`{!isEmpty(agents) && <RecentCounterparties />}`

vs putting it in the component itself

`if (isEmpty(agents)) return null`

@JettTech your question about having the header visible even if there are no agents to list is good one, specific to this component, and a UX question we should throw back to C&Z

src/holofuel/pages/CreateRequest/CreateRequest.test.js
robbiecarlton: right. The two test slightly different things. Your test is slightly more determined than mine, in that yours tests that agent1 appears in the first row, where mine just tests that it appears in one of the rows. I agree it doesn't really matter for this test.

src/holofuel/pages/CreateOffer/CreateOffer.js
robbiecarlton: Good call.

src/holofuel/pages/CreateOffer/CreateOffer.test.js
robbiecarlton: Good catch, removed.

src/holofuel/pages/CreateRequest/CreateRequest.test.js
robbiecarlton: Good catch, removed.

JettTech (approved):
Ship it! :package: :ship:

PR #115, opened 2019-11-15
6369 - As a HoloFuel User, when pressing confirm in an action Confirmation Modal, I am redirected to the Inbox page with the correct Filter

JettTech:


src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -368,6 +369,7 @@ export function ConfirmationModal ({ transaction, handleClose, declineTransactio
 
   const onYes = () => {
     actionHook(actionParams)
+    if (actionHook === declineTransaction) setInboxView(VIEW['recent'])
```
robbiecarlton: i would make this conditional on `action`, not `actionHook`. 
ie
`if (action === 'decline') setInboxView(VIEW['recent'])`

Couple of principles inform this for me
1) single source of truth. `action === 'decline'` is already how the component knows which action it's taking, so that should be our test everywhere we're asking that question
2) it's best to avoid comparing object references when possible, and especially function references.

robbiecarlton (changes requested):
One change requested in a comment.

I also noticed when testing this manually that `actionsVisible` isn't reset to empty when you take the action. I know this isn't super important because in practice the row won't be there. But I think it's good just to clean up state and have it be as accurate a representation of what we want the UI to look like as possible.

So I think this means passing `setActionsVisible` into ConfirmationModal, and calling it with `null` in `onYes`

src/holofuel/pages/Inbox/Inbox.js
JettTech: Solid points; I agree with this logic. :raised_hands:

robbiecarlton (approved):
Approved, good to merge once Inbox has been merged

PR #119, opened 2019-11-18
5007 - Implement v0.1.3 UI design to HP Admin - Hosting

robbiecarlton:


src/components/UIButton/UIButton.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
```
robbiecarlton, line 1: I created a new Button component because the current one has a bunch of cruft in it that's not relevant. We should use this button for the rounded green and white Hylo style buttons in the designs.

JettTech: Should we not just modify or just completely replace the other Button component, do avoid any name/reference confusion? I do agree that is prob a bit outdated. :thinking:

JettTech (approved):
LGTM! I made a comment that we actually eneded up discussing yesterday, but am keeping here just to keep tabs on our thinking/convo. :+1:

PR #121, opened 2019-11-22
5005 - v0.1.3 UI design to HP Admin - Earnings

robbiecarlton:


src/data-interfaces/mockEarningsData.js
```diff
@@ -0,0 +1,156 @@
+import { UNITS } from 'models/HostPricing'
```
JettTech, line 1: Thanks for extracting this! :100:

JettTech (approved):
LGTM!  Beautiful implementation of the graphs. :+1:  Side note: I would like to look into more ways to test the graphs... but I think we're good for now.

PR #123, opened 2019-11-25
hp-admin-ui -> hp-admin

lukateras:


default.nix
```diff
@@ -68,8 +68,8 @@ in
 
 {
 
-  hp-admin-ui = stdenv.mkDerivation rec {
-    name = "hp-admin-ui";
+  hp-admin = stdenv.mkDerivation rec {
```
zo-el, line 71: @yegortimoshenko, I named this `hp-admin-ui` to be specific. Also, We have `holofuel-ui` in the same repo, So its just following a certain pattern.

I do not have anything against this naming though. 

Note: If we choose to go with this we have to update the derivation name in `holo-nixpkgs` as well

lukateras:
Sorry. I mostly made this change to test if Hydra integration works or not.

PR #124, opened 2019-11-26
6609 - BUG: On latest branch of develop the Inbox page crashes.

robbiecarlton:


src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -46,38 +46,25 @@ function useDecline () {
   })
 }
 
-function useFetchCounterparties () {
+function useTransactionsWithCounterparties () {
```
robbiecarlton, line 49: I changed the way this worked actually back to how you were originally doing it.

That's because the update caching method wasn't actually working how we thought. Remember how

`updatedTransactions = transactions.map(transaction => { Object.assign(transaction.counterparty, agent); return transaction })`

worked to update the transactions, but

`updatedTransactions = transactions.map(transaction => ({
  ...transaction,
  counterparty: agent
})`

didn't?

That's because we were actually relying on the *sideeffect* of `Object.assign`, not the return value. We were modifying the objects in cache directly, not using writeCache. We don't want to do this because it's unspecified behavior. So I switched to the simpler thing of just returning updated versions of the transaction lists.

(none of this was directly to do with the bug, just ended up cleaning this up while fixing the bug).

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -46,38 +46,25 @@ function useDecline () {
...
-  if (holofuelInboxCounterparties) {
-    const filterTransactionsByAgentId = (agent, txListType) => txListType.filter(transaction => transaction.counterparty.id === agent.id)
-    const updateTxListCounterparties = (txListType, counterpartyList) => counterpartyList.map(agent => {
-      const matchingTx = filterTransactionsByAgentId(agent, txListType)
-      return matchingTx.map(transaction => { Object.assign(transaction.counterparty, agent); return transaction })
-    })
-
-    // Cache Write/Update for holofuelActionableTransactions
-    const reflexiveActionableTxList = updateTxListCounterparties(holofuelActionableTransactions, [whoami])
-    const newActionableTxList = flatten((updateTxListCounterparties(holofuelActionableTransactions, holofuelInboxCounterparties)).concat(reflexiveActionableTxList))
```
robbiecarlton, line 64: this line was the cause of the bug.

If you have sent a transaction to yourself, both `holofuelActionableTransactions` and `reflexiveActionableTxList` have a copy of the transaction. So then when you write the new list to the cache, it's double in length, and so this query reruns, and now they each have 2 copies of the same transaction in them and so `newActionableTxList` has four copies, and the whole process feeds back on itself, with the transaction list doubling in size every time.

The fix for this is a little hidden in the updated code because I changed a lot of how this function works, but the `uniqBy` in line 60 prevents the same thing happening in the new code.

The very simplest version of the bug fix would be to add a `uniqBy` to line 64 in the old code.

robbiecarlton:
@JettTech - when you review this, will you run the code and make sure that Inbox page doesn't crash anymore. Pretty confident I caught the bug, but as I wasn't the one who found it, it's possible I missed something.

src/holofuel/pages/Inbox/Inbox.js
JettTech: Good catch!  This was definitely the culprit! :tada:

src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -29,45 +30,33 @@ function useCancel () {
...
-      query: HolofuelWaitingTransactionsQuery,
-      data: {
-        holofuelWaitingTransactions: newWaitingTxList
-      }
-    })
+
+  const updateCounterparties = (transactions, counterparties) => transactions.map(transaction => ({
+    ...transaction,
+    counterparty: counterparties.find(counterparty => counterparty.id === transaction.counterparty.id) || transaction.counterparty
+  }))
```
JettTech, line 42: Nice way of reworking the code to leverage the calls we are already making.  I'm glad to see we didn't fully return to an external renderNickname component.

Also, I know the fetch for counterparty nicknames has now been re-worked closer to its original form, but during optimization stages, let's consider Apollo's Read/Write Fragments (https://www.apollographql.com/docs/react/data/local-state/#writefragment-and-readfragment), which can be triggered them from within the holofuelCounterparty query resolver.  This could be a good way to still populate the Transaction schema with the counterparty nickname while reducing call frequency to `getCounterparty`.

src/holofuel/pages/Inbox/Inbox.js
JettTech: Interesting thought.  Recently, when reading through the Apollo docs to consider best solutions for PT Ticket #1844 and found that Apollo recommends the use of 'Object.assign' when using Subscriptions' `updateQuery` option to update cached data from Queries.  I'd love to contrast these approaches together to understand how the `updateQuery` option handles cache writing differently, such that the use 'Object.assign' is a preferred implementation.

Nonetheless, good call to avoid any side effects. :+1:

JettTech (approved):
Nice work overall.  Good idea to return to what we know works for now. :shipit:

src/holofuel/pages/TransactionHistory/TransactionHistory.js
robbiecarlton: sounds good

PR #127, opened 2019-12-02
Updated Instance ID

zo-el:
Nicknames we updated to are as follows:

- holo-hosting-app
- happ-store
- holofuel

> Note: This update would work only for one agent as discussed.
> This will be solved when we update the current holochain version

### DIFF
--- a/src/holochainClient.js
+++ b/src/holochainClient.js
@@ -21,17 +21,14 @@ export const MOCK_INDIVIDUAL_DNAS = {
 export const HOLOCHAIN_LOGGING = true && process.env.NODE_ENV !== 'test'
 let holochainClient
 
-// const agentId = process.env.REACT_APP_AGENT_ID
-// export function conductorInstanceId (instanceId) {
-//   const formatInstanceId = instanceId => agentId + '::<happ_id>-' + instanceId
-
-//   return {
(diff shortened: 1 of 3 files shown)


src/holochainClient.js
```diff
@@ -21,16 +21,12 @@ export const MOCK_INDIVIDUAL_DNAS = {
...
-
+export function conductorInstanceIdbyDnaAlias (instanceId) {
   return {
-    hylo: formatInstanceId('hylo'),
-    'happ-store': formatInstanceId('happ-store'),
-    hha: formatInstanceId('holo-hosting-app'),
-    holofuel: formatInstanceId('holofuel')
+    hylo: 'hylo',
+    'happ-store': 'happ-store',
+    hha: 'holo-hosting-app',
```
JettTech, line 28: This dictionary lookup only exists because of the hha to holo-hosting-app conversion.  It would be nice to either update nix to expect the 'hha' alias, or to update every appearance of 'hha' to 'holo-hosting-app' in hp-admin.  Happy to hear votes for either direction. For now, this works.

JettTech (approved):
LGTM.

robbiecarlton (approved):

PR #129, opened 2019-12-03
update the holo nixpkgs to point to holochain v0.0.39-alpha4

zo-el:


JettTech (approved):
Please and thank you! :100: :tada:

PR #130, opened 2019-12-04
1844 data polling bug & 6797 production vs develop reference to `hc-web-client connect()`

JettTech:


package.json
```diff
@@ -1,12 +1,13 @@
...
   "name": "hp-admin",
-  "version": "0.1.0",
+  "version": "0.2.1",
   "private": true,
   "scripts": {
     "start": "rm -rf ./node_modules/.cache/babel-loader && node scripts/start.js",
     "start:holofuel": "REACT_APP_HOLOFUEL_APP=true npm start",
     "build": "node scripts/build.js",
     "build:holofuel": "REACT_APP_HOLOFUEL_APP=true node scripts/build.js",
+    "hc:start-manual": "holochain -c conductor-config-manuals.toml",
```
robbiecarlton: `conductor-config-manuals.toml` should be `conductor-config-manual.toml`

pkgs.nix
```diff
@@ -1,4 +1,4 @@
 import (fetchTarball {
-  url = "https://github.com/Holo-Host/holo-nixpkgs/archive/6c5491909b4fa1a4a1d7864fa204c1e5f28a6a1f.tar.gz";
-  sha256 = "1r5d3dqfi20xxjnbzwr1i5dsr7blc1q55dha29cfvi81dz52fmqr";
+  url = "https://github.com/Holo-Host/holo-nixpkgs/archive/c4ed4fbf07d74cbbcaec5a338e062c3114b7a825.tar.gz";
+  sha256 = "0sqjcqj6c862lilxb4siwm1rxi2bp8042r1bh1k4xc7gi5c2igvf";
```
robbiecarlton: this bumps us to v39?

scripts/holochain-client.js
```diff
@@ -8,7 +8,6 @@ async function initAndGetHolochainClient () {
   if (holochainClient) return holochainClient
   try {
     holochainClient = await hcWebClientConnect({
-      url: process.env.REACT_APP_DNA_INTERFACE_URL,
```
robbiecarlton, line 11: why remove this env var?

src/holochainClient.js
```diff
@@ -34,7 +34,7 @@ async function initAndGetHolochainClient () {
   if (holochainClient) return holochainClient
   try {
     holochainClient = await hcWebClientConnect({
-      url: process.env.REACT_APP_DNA_INTERFACE_URL,
+      url: 'ws://localhost:3400/',
```
robbiecarlton: why replace with hardcoded version?

src/holofuel/components/SideMenu/SideMenu.js
```diff
@@ -106,10 +106,10 @@ export function SideMenu ({
       <footer styleName='footer'>
         <ul styleName='footer-list'>
           <li>
-            <Link to='/tos' styleName='footer-link'>View Disclaimer</Link>
+            <Link to='#/' styleName='footer-link'>View Disclaimer</Link>
```
robbiecarlton: does this mean we can remove the `TOS` page entirely?

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -17,9 +17,9 @@ export function PrimaryLayout ({
   children,
   headerProps = {}
 }) {
-  const { data: { holofuelActionableTransactions: transactions = [] } = {} } = useQuery(HolofuelActionableTransactionsQuery)
+  const { data: { holofuelActionableTransactions: transactions = [] } = {} } = useQuery(HolofuelActionableTransactionsQuery, { fetchPolicy: 'network-only' })
```
robbiecarlton, line 20: This did the trick?

It's definitely a bummer and we want to keep looking at other ways to do this as we want to get the benefit of caching.

robbiecarlton (changes requested):

src/holochainClient.js
zo-el: Can we implement this in this way:
```
const url = process.env.NODE_ENV === 'production' ? undefined : process.env.REACT_APP_DNA_INTERFACE_URL
```

zo-el (changes requested):
We need to deal with automating the `url` var

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
zo-el: @robbiecarlton Honestly, I don't see any other way other than this implementation from @JettTech. 
Because we need to pull this from the network.
Correct me if I am wrong, but the only way this is solved is when we implement `signals` in the Holofuel DNA so that we can use some sort of `subscription` method

src/holofuel/components/SideMenu/SideMenu.js
JettTech: The TOS has been removed from the router, so I changed this link to a placeholder to avoid erroring out.  Has the design changed to remove this entirely? Do we have clarity on whether this will be a page internal to our app?  I would like to defer to the designs to answer this question.

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: @robbiecarlton - True, as we discussed previously, forcing network calls is not ideal, but does work for now.  As we've also discussed, and as @zo-el mentioned here, I do think we can revisit and reform this approach once we implement signals and subscriptions.  That said, I am still open to implementing a more optimal solution, if you know one. 

@zo-el - I agree with the idea of optimizing our calls with subscriptions and signals once signals are implemented in the hf dna. Is there any movement toward this in Holo Central?

package.json
JettTech: Good catch! :)

pkgs.nix
JettTech: Yes, this was tied to v39 and is dependent on PR#129, so we should merge that first. :+1:

scripts/holochain-client.js
JettTech: After preparing our build for holoscape and building around recommended hc patterns, I highly recommend referencing logic that determines whether we're in a development or production environment and assigns the url value accordingly.  So, this is a cleanup task.  I will turn this into a logical statement. :+1:

src/holochainClient.js
JettTech: Same thought/reason as above.  I meant to clean up this hard-coded vars, so thanks for that catch!  @zo-el - I like the logical statement you proposed, and my intention is to implement this or something similar to leverage the benefit of a env var for testing, but following the 'hc-web-client' expected behavior for production environments for best coordination with the conductor and other hc apps.

src/holofuel/components/SideMenu/SideMenu.js
JettTech: Ok. So I checked and it looks like TOS is still apart of the sidebar menu, but disclaimer is not and a link to 'Help' has been since added.  I made the updates accordingly. :+1:

src/holochainClient.js
robbiecarlton: this just seems redundant

`const url = process.env.REACT_APP_DNA_INTERFACE_URL`

is the same as your suggestion @zo-el. This is what env variables are for. Set the variable how you need it in the specific environment.

src/holochainClient.js
```diff
@@ -34,7 +34,7 @@ async function initAndGetHolochainClient () {
   if (holochainClient) return holochainClient
   try {
     holochainClient = await hcWebClientConnect({
-      url: process.env.REACT_APP_DNA_INTERFACE_URL,
+      url: process.env.NODE_ENV === 'production' ? undefined : process.env.REACT_APP_DNA_INTERFACE_URL,
```
robbiecarlton, line 37: this is the same as `url: process.env.REACT_APP_DNA_INTERFACE_URL`

I don't understand why we're doing this. Just let the environment determine it.

robbiecarlton (approved):
I feel differently to you and Joel about specifying what the 'url' param should be in production.

It just removes flexibility and doesn't gain anything. What good thing comes from us restraining the value of the url when in a production environment?

That being said, don't let it block the PR.

src/holochainClient.js
zo-el: @robbiecarlton Yes, but this would mean that Everytime you have to make a build you will have to remove your env var from your .env file.
But if u use the NODE_ENV you do not have to deal with this issue. This helps in automation.
Especially when u have to run test as well as make a build, which we are doing on Hydra.

robbiecarlton:
we can handle that by having separate .env files for production and
development

zo-el:
I don't think that is standard. Also it doesn't solve issues with CI's

robbiecarlton:
It’s definitely normal practice. it’s built into create-react-app. And I
think you can have one for .env.test for CI

https://dev.to/jam3/managing-env-variables-for-provisional-builds-h37

zo-el:
I think the CI should not be testing things out on different environment settings. 
I think the testing should be done in a `development` setting and the build should be done in `production` settings.
Also in CI's, we do not have .env files. its easier to just set the env var without the `.env`, which makes using `.env.test` not really optimal.
I did look through the above link and I don't see a need for the third environment, because there is nothing special about the CI environment. And we would need another dependency to implement it.
Also, we solve all of these issues just by using `NODE_ENV`. 
The `NODE_ENV` is used in create_react_app anyway in the react-scripts.

Is there a reason why you prefer not using this variable? 
That would probably clear things out for me.

robbiecarlton:
It’s just redundant. The whole point of an env variable is to set the
variable to what you need in a specific environment. This is a canonical
use case of an env variable. So instead we’re just jury rigging an env
variable when we could be using the mechanism which is designed for this
purpose.

So for CI we set the variable to whatever we need, in production we set it
and in development we set it. Done. The code doesn’t care because the code
doesn’t need to know.

zo-el:
I don't understand what is redundant. Can you explain?

zo-el (approved):

PR #131, opened 2019-12-05
0783 - Implement latest hpos api

robbiecarlton:


src/data-interfaces/HposInterface.js
```diff
@@ -1,6 +1,15 @@
 import axios from 'axios'
 import mockCallHpos from 'mock-dnas/mockCallHpos'
 
+// these two functions are stubs until we have the crypto library
+function payloadSignature (authKey, method, request, body) {
+  return 'signature'
+}
+
+function hashResponseBody (body) {
+  return 'hash of body'
```
lukateras, line 10: https://www.npmjs.com/package/fast-json-stable-stringify

```js
import jsonStableStringify from 'fast-json-stable-stringify'

const hashResponseBody = async (data) => {
  const dataBytes = Buffer.from(jsonStableStringify(data))
  const hashBytes = await crypto.subtle.digest('SHA-512', dataBytes)

  return Buffer.from(hashBytes).toString('base64')
}
```

robbiecarlton: great, thanks! do we have the `crypto` library available yet? @zo-el told me that I should use the specific library you guys are working on.

lukateras: For signatures, we have library and API that's required, but using it requires WASM scaffolding. Should I send a PR to this repo that would add that in?

PR #132, opened 2019-12-06
1170(Update HF Interface with new DNA Cancel & Decline endpoints), 0055(bug inbox offer), and 4770 (bug transaction restriction)

JettTech:


zo-el (approved):

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -161,7 +161,7 @@ const HoloFuelDnaInterface = {
   user: {
     get: async () => {
       const result = await createZomeCall('transactions/whoami')()
-      if (result.error) throw new Error('There was an error locating the current holofuel agent nickname. ERROR: ', result.error)
+      if (result.Err) return console.error('There was an error locating the current holofuel agent nickname. ERROR: ', result.Err)
```
robbiecarlton: throw error here

src/graphql-server/schema.graphql
```diff
@@ -35,6 +35,7 @@ type HostPricing {
 type HolofuelUser {
   id: ID
   nickname: String
+  error: Boolean
```
robbiecarlton: rename to `notFound`

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -170,7 +170,13 @@ const HoloFuelDnaInterface = {
     },
     getCounterparty: async ({ agentId }) => {
       const result = await createZomeCall('transactions/whois')({ agents: agentId })
-      if (result.error || !result[0].Ok) throw new Error('There was an error locating the counterparty agent nickname. ERROR: ', result.error)
+      if (result.Err || !result[0].Ok) {
+        console.error('There was an error locating the current holofuel agent nickname. ERROR: ', result.Err || result[0].Err)
+        return {
+          id: agentId,
+          error: true
```
robbiecarlton: rename to `notFound`

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -170,7 +170,13 @@ const HoloFuelDnaInterface = {
     },
     getCounterparty: async ({ agentId }) => {
       const result = await createZomeCall('transactions/whois')({ agents: agentId })
-      if (result.error || !result[0].Ok) throw new Error('There was an error locating the counterparty agent nickname. ERROR: ', result.error)
+      if (result.Err || !result[0].Ok) {
+        console.error('There was an error locating the current holofuel agent nickname. ERROR: ', result.Err || result[0].Err)
```

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -38,17 +38,26 @@ function useOfferMutation () {
...
 
   useEffect(() => {
     setCounterpartyNick(presentAgentId(counterpartyId))
   }, [counterpartyId])
 
-  const { register, handleSubmit, errors, setValue: setFormValue } = useForm({ validationSchema: FormValidationSchema })
+  const { register, handleSubmit, errors, setValue: setFormValue, getValues } = useForm({ validationSchema: FormValidationSchema })
+  const formValues = getValues()
+
+  if (errorMessage) {
```
robbiecarlton:
```
useEffect(() => {
  if (errorMessage) {
      newMessage(errorMessage)
      setErrorMessage('')
  }
}, [errorMessage, setErrorMessage])
```

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -38,17 +38,26 @@ function useOfferMutation () {
...
   const { data: { holofuelHistoryCounterparties: agents } = {} } = useQuery(HolofuelHistoryCounterpartiesQuery)
-
   const createOffer = useOfferMutation()
 
+  const { newMessage } = useFlashMessageContext()
+
   const [counterpartyId, setCounterpartyId] = useState('')
   const [counterpartyNick, setCounterpartyNick] = useState('')
+  const [errorMessage, setErrorMessage] = useState('')
+  const [disabled, setDisabled] = useState(true)
```
robbiecarlton: `rename` disabled `counterpartyNotFound`

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -131,17 +144,18 @@ export default function CreateOffer ({ history: { push } }) {
...
         placeholder='Notes'
         ref={register} />
+
       <RecentCounterparties
         styleName='recent-counterparties'
         agents={agents}
         selectedAgentId={counterpartyId}
         selectAgent={selectAgent} />
-      <Button type='submit' wide variant='secondary' styleName='send-button'>Send</Button>
+      <Button type='submit' wide variant='secondary' styleName='send-button' disabled={counterpartyId.length === AGENT_ID_LENGTH ? disabled : true}>Send</Button>
```
robbiecarlton: `disabled = {counterpartyNotFound || counterpartyId.length !== AGENT_ID_LENGTH}`

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -91,7 +98,13 @@ export default function CreateOffer ({ history: { push } }) {
...
           {counterpartyId.length === AGENT_ID_LENGTH && <HashIcon hash={counterpartyId} size={26} styleName='hash-icon' />}
           {counterpartyId.length === AGENT_ID_LENGTH && <h4 data-testid='counterparty-nickname'>
-            <RenderNickname agentId={counterpartyId} setCounterpartyNick={setCounterpartyNick} />
+            <RenderNickname
+              agentId={counterpartyId}
+              setCounterpartyNick={setCounterpartyNick}
+              setErrorMessage={setErrorMessage}
+              setDisabled={setDisabled}
+              disabled={disabled}
+              formValues={formValues} />
```
robbiecarlton: let's just pass in `counterpartyId` instead of all `formValues`

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -150,6 +164,24 @@ export function RenderNickname ({ agentId, setCounterpartyNick }) {
     setCounterpartyNick(nickname)
   }, [setCounterpartyNick, nickname])
 
+  let errorMessage = null
```
robbiecarlton:
```
useEffect(() => {
  if (!nickname && !loading) {
    setErrorMessage('This HoloFuel Peer is currently unable to be located in the network. \n Please verify the hash, ensure your HoloFuel Peer is online, and try again after a few minutes.')
    setDisabled(true)
  } else if (formValues.counterpartyId && formValues.counterpartyId.length === AGENT_ID_LENGTH) {
    setDisabled(false)
  }
})
```

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -131,17 +144,18 @@ export default function CreateOffer ({ history: { push } }) {
...
         selectAgent={selectAgent} />
-      <Button type='submit' wide variant='secondary' styleName='send-button'>Send</Button>
+      <Button type='submit' wide variant='secondary' styleName='send-button' disabled={counterpartyId.length === AGENT_ID_LENGTH ? disabled : true}>Send</Button>
     </form>
   </PrimaryLayout>
 }
 
-export function RenderNickname ({ agentId, setCounterpartyNick }) {
+export function RenderNickname ({ agentId, setCounterpartyNick, setErrorMessage, setDisabled, disabled, formValues }) {
   const { loading, error, data: { holofuelCounterparty = {} } = {} } = useQuery(HolofuelCounterpartyQuery, {
```
robbiecarlton: call this `error: queryError`

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -150,6 +164,24 @@ export function RenderNickname ({ agentId, setCounterpartyNick }) {
     setCounterpartyNick(nickname)
   }, [setCounterpartyNick, nickname])
 
+  let errorMessage = null
+  if (!nickname && !loading) {
```
robbiecarlton: don't test on `nickname`, test on `holofuelCounterparty.notFound`

src/holofuel/pages/CreateOffer/CreateOffer.js
```diff
@@ -163,6 +195,8 @@ export function RenderNickname ({ agentId, setCounterpartyNick }) {
     </>
   }
 
-  if (error || !nickname) return <>No nickname available.</>
+  if (error || !nickname) {
```
robbiecarlton: remove curly braces

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -46,6 +48,14 @@ function useDecline () {
   })
 }
 
+function useCounterparty (agentId) {
+  const { loading, data: { holofuelCounterparty = {} } = {} } = useQuery(HolofuelCounterpartyQuery, {
+    variables: { agentId }
+  })
+  if (loading) return null
```
robbiecarlton: I don't think you don't need line 55

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -325,10 +338,21 @@ function NewTransactionModal ({ handleClose, toggleModal }) {
   </Modal>
 }
 
-export function ConfirmationModal ({ transaction, handleClose, declineTransaction, payTransaction }) {
-  if (!transaction) return null
+export function ConfirmationModal ({ transaction, handleClose, declineTransaction, payTransaction, setDisabled, disabled }) {
```
robbiecarlton: `transaction = {}`

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -325,10 +338,21 @@ function NewTransactionModal ({ handleClose, toggleModal }) {
...
+  if (counterpartyConfirmed) {
+    setDisabled(false)
+    const errorMessage = 'This HoloFuel Peer is currently unable to be located in the network. \n Please confirm your HoloFuel Peer is online, and try again after a few minutes.'
+    console.log(' should show error message? =>>> counterpartyConfirmed.nickname !== counterparty.nickname : ', counterpartyConfirmed.nickname !== counterparty.nickname)
+    if (counterpartyConfirmed.nickname !== counterparty.nickname) {
+      setDisabled(true)
+      newMessage(errorMessage)
+    }
+  }
+
```
robbiecarlton, line 393: if (!transaction) return null

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -46,6 +48,14 @@ function useDecline () {
   })
 }
 
+function useCounterparty (agentId) {
+  const { loading, data: { holofuelCounterparty = {} } = {} } = useQuery(HolofuelCounterpartyQuery, {
+    variables: { agentId }
+  })
+  if (loading) return null
+  return holofuelCounterparty
```
robbiecarlton: return { holofuelCounterparty, loading }

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -325,10 +338,21 @@ function NewTransactionModal ({ handleClose, toggleModal }) {
   </Modal>
 }
 
-export function ConfirmationModal ({ transaction, handleClose, declineTransaction, payTransaction }) {
-  if (!transaction) return null
+export function ConfirmationModal ({ transaction, handleClose, declineTransaction, payTransaction, setDisabled, disabled }) {
+  const { newMessage } = useFlashMessageContext()
   const { id, counterparty, amount, type, action } = transaction
 
+  const counterpartyConfirmed = useCounterparty(counterparty.id)
```
robbiecarlton:
```
const { holofuelCounterparty, loading }

useEffect(() => {
  if (!loading) {
    if (holofuelCounterparty.notFound) {
      setDisabled(true)
      newMessage('This HoloFuel Peer is currently unable to be located in the network. \n Please confirm your HoloFuel Peer is online, and try again after a few minutes.')
    } else {
      setDisabled(false)
    }
  }
})

File: src/mock-dnas/holofuel.js [hunk 18]
```diff
@@ -325,7 +325,7 @@ const agentArray = [{
   }
 }]
 
-const whois = (agentId) => agentArray.find(agent => agent.Ok.agent_id.pub_sign_key === agentId) || { error: 'No agent was found by this id.' }
+const whois = (agentId) => agentArray.find(agent => agent.Ok.agent_id.pub_sign_key === agentId) || { Err: 'No agent was found by this id.' }
```
robbiecarlton: remove parens

robbiecarlton (changes requested):

src/holofuel/pages/CreateOffer/CreateOffer.js
JettTech: So I need to leave the ternary operator for the case wherein the  counterpartyId length is less than full amount and the counterpartyNotFound hasn't been run yet (as it is only trigged once the counterpartyId length is satisfied)

src/data-interfaces/__mocks__/HoloFuelInterface.js
```diff
@@ -23,8 +23,8 @@ const mockHoloFuelDnaInterface = {
     allCompleted: jest.fn(() => ['mockCompleteTransactionOne', 'mockCompleteTransactionTwo']),
     allActionable: jest.fn(() => ['mockActionableTransactionOne', 'mockActionableTransactionTwo']),
     allWaiting: jest.fn(() => ['mockWaitingTransactionOne', 'mockWaitingTransactionTwo']),
-    decline: jest.fn(transactionId => successfulTransactionResponse({ transactionId, status: STATUS.declined, type: TYPE.offer })), // type is hard-coded, but could be either
-    cancel: jest.fn(transactionId => successfulTransactionResponse({ transactionId, status: STATUS.cancelled, type: TYPE.outgoing })) // type is hard-coded, but could be either
+    decline: jest.fn(transactionId => successfulTransactionResponse({ transactionId, status: STATUS.declined, type: TYPE.request })),
+    cancel: jest.fn(transactionId => successfulTransactionResponse({ transactionId, status: STATUS.canceled, type: TYPE.offer })) // type is hard-coded, but could be eitherfor *syncnhronous* transactions
```
robbiecarlton: not sure what this comment means.

package.json
```diff
@@ -4,10 +4,14 @@
   "private": true,
   "scripts": {
     "start": "rm -rf ./node_modules/.cache/babel-loader && node scripts/start.js",
+    "start:holofuel-agent-1": "rm -rf ./node_modules/.cache/babel-loader && REACT_APP_HOLOFUEL_APP=true PORT=3001 REACT_APP_DNA_INTERFACE_URL=ws://localhost:3400 node scripts/start.js",
```
robbiecarlton: 👍

package.json
```diff
@@ -4,10 +4,14 @@
...
+    "start:holofuel-agent-2": "rm -rf ./node_modules/.cache/babel-loader && REACT_APP_HOLOFUEL_APP=true PORT=3002 REACT_APP_DNA_INTERFACE_URL=ws://localhost:3500 REACT_APP_HPOS_URL=3200 node scripts/start.js",
     "start:holofuel": "REACT_APP_HOLOFUEL_APP=true npm start",
     "build": "node scripts/build.js",
     "build:holofuel": "REACT_APP_HOLOFUEL_APP=true node scripts/build.js",
     "hc:start-manual": "holochain -c conductor-config-manual.toml",
```
robbiecarlton: let's rename this `hc:start-manual-1` by analogy with the `start-holofuel-agent-*` commands

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -105,24 +108,51 @@ const presentCheque = ({ origin, event, stateDirection, eventTimestamp, fees, pr
   }
 }
 
-function presentPendingRequest (transaction) {
-  const { event, provenance } = transaction
+const presentDeclinedTransaction = declinedTx => {
+  if (!declinedTx[2]) return console.log('declinedTx[2] is UNDEFINED : ', declinedTx)
```
robbiecarlton: should throw.

we want to keep `console.log` as clean as possible. Fine for debugging, but in `develop` we almost never want to be calling `console.log`. And especially when something unexpected is happening.

Specifically in this case:

presentCanceledTransaction should always return an object in the shape of our graphql Transaction type. So when you return `undefined` (which is the return value of console.log), graphql will complain and throw an error anyway and it will be harder to debug because the error will be further from the source. 

In general, we want to follow the principle `fail early, fail loudly`

http://oncodingstyle.blogspot.com/2008/10/fail-early-fail-loudly.html

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -195,66 +237,140 @@ const HoloFuelDnaInterface = {
...
+      return noDuplicateIds.filter(tx => tx.status === 'completed').sort((a, b) => a.timestamp > b.timestamp ? -1 : 1)
     },
     allActionable: async () => {
-      const { requests, promises } = await createZomeCall('transactions/list_pending')()
-      const actionableTransactions = requests.map(presentPendingRequest).concat(promises.map(presentPendingOffer))
-      return actionableTransactions.sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
+      const { requests, promises, declined, canceled } = await createZomeCall('transactions/list_pending')()
+      const actionableTransactions = await requests.map(r => presentPendingRequest(r)).concat(promises.map(p => presentPendingOffer(p))).concat(declined.map(presentDeclinedTransaction)).concat(canceled.map(presentCanceledTransaction))
+
+      console.log('ALL ACTIONABLE TRANSACTIONS : ', actionableTransactions.sort((a, b) => a.timestamp > b.timestamp ? -1 : 1))
```

src/utils/index.js
```diff
@@ -15,6 +15,7 @@ export async function promiseMap (array, fn) {
 }
 
 export function sliceHash (hashString = '', desiredLength = 6) {
+  // console.log(parseInt(hashString) || hashString === '' ? '' : `hashString: ${hashString}`)
```
robbiecarlton: remove comment

src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -32,13 +32,17 @@ function useCancel () {
...
-  const { data: { holofuelHistoryCounterparties = [] } = {} } = useQuery(HolofuelHistoryCounterpartiesQuery)
+  const { data: { holofuelHistoryCounterparties = [] } = {} } = useQuery(HolofuelHistoryCounterpartiesQuery, { fetchPolicy: 'network-only' })
   const { data: { holofuelCompletedTransactions = [] } = {} } = useQuery(HolofuelCompletedTransactionsQuery, { fetchPolicy: 'network-only' })
   const { data: { holofuelWaitingTransactions = [] } = {} } = useQuery(HolofuelWaitingTransactionsQuery, { fetchPolicy: 'network-only' })
 
   const updateCounterparties = (transactions, counterparties) => transactions.map(transaction => ({
     ...transaction,
-    counterparty: counterparties.find(counterparty => counterparty.id === transaction.counterparty.id) || transaction.counterparty
+    counterparty: counterparties.find(counterparty => {
+      if (transactions.counterparty != null) {
```
robbiecarlton: I would just use lodash `get` instead of checking for null here. So line 41 would be 
```
counterparty: counterparties.find(counterparty => counterparty.id === get('counterparty.id', transaction)) || transaction.counterparty
```

same thing but less reading

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -359,8 +418,17 @@ export function ConfirmationModal ({ transaction, handleClose, declineTransactio
...
+      contentLabel = `Cancel ${type}?`
+      actionParams = { id }
+      actionHook = refundTransaction
+      message = <div styleName='modal-text' data-testid='modal-message'>Cancel your declined {type} of <span styleName='modal-amount'>{presentHolofuelAmount(amount)} HF</span> {type === TYPE.offer ? 'to' : 'from'} <span styleName='counterparty'> {counterparty.nickname || presentAgentId(counterparty.id)}</span>?<br /><br /><div styleName='modal-note-text'>Note: Canceling will refund your balance by the outstanding amount.</div></div>
+      break
+    }
     default:
-      throw new Error('Error: Transaction action was not matched with a modal action. Current transaction action : ', action)
+      // NB: action === undefined when first loading page && no transaction is yet passed in
+      if (action === undefined) break
```
robbiecarlton, line 442: good catch 👍

src/holofuel/pages/Inbox/Inbox-backup.js
```diff
@@ -0,0 +1,403 @@
+import React, { useState } from 'react'
```
robbiecarlton: I think you can remove this file now.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -39,14 +41,34 @@ function useOffer () {
 
 function useDecline () {
   const [decline] = useMutation(HolofuelDeclineMutation)
-  return ({ id }) => decline({
+  return ({ id }) => {
+    console.log('THE TRANSATION - id in useDecline', id)
```
robbiecarlton: remove log

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -61,10 +83,20 @@ function useTransactionsWithCounterparties () {
   const allCounterparties = uniqBy('id', holofuelInboxCounterparties.concat([whoami]))
 
   const updatedActionableTransactions = updateCounterparties(holofuelActionableTransactions, allCounterparties)
-  const updatedNonPendingTransactions = updateCounterparties(holofuelNonPendingTransactions, allCounterparties)
+  console.log('updatedActionableTransactions', updatedActionableTransactions)
```
robbiecarlton: remove log

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -61,10 +83,20 @@ function useTransactionsWithCounterparties () {
...
+
+  const updatedActionableWOCanceledOffers = updatedActionableTransactions.filter(actionableTx => (actionableTx.status !== STATUS.canceled) && (actionableTx.type !== TYPE.request))
+  console.log('updatedActionableWOCanceledOffers', updatedActionableWOCanceledOffers)
```
robbiecarlton: remove log

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -61,10 +83,20 @@ function useTransactionsWithCounterparties () {
...
+
+  const updatedCanceledTransactions = updatedActionableTransactions.filter(actionableTx => actionableTx.status === STATUS.canceled)
+  const updatedDeclinedTransactions = updatedActionableTransactions.filter(actionableTx => actionableTx.status === STATUS.declined)
+  const updatedNonPendingTransactions = updateCounterparties(holofuelNonPendingTransactions, allCounterparties).concat(updatedCanceledTransactions).concat(updatedDeclinedTransactions)
+
+  // console.log('updatedActionableWithoutCanceled : ', updatedActionableWOCanceledOffers)
```
robbiecarlton: remove comments

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -61,10 +83,20 @@ function useTransactionsWithCounterparties () {
...
+  // console.log('updatedNonPendingTransactions : ', updatedNonPendingTransactions)
 
   return {
-    actionableTransactions: updatedActionableTransactions,
+    actionableTransactions: updatedActionableWOCanceledOffers,
```
robbiecarlton, line 104: I think this implementation is good for now, but this function is starting to get a little too bloated to me. I think there's an opportunity for a better abstraction here. Probably new queries, that move this filtering work out of the front end.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -39,14 +41,34 @@ function useOffer () {
...
+    decline({
+      variables: { transactionId: id },
+      refetchQueries: [{
+        query: HolofuelActionableTransactionsQuery
+      }]
+    })
+  }
+}
+
+function useRecoverFunds () {
```
robbiecarlton: I think `useRefund` makes more sense given how it's used and by analogy with `useDecline`.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -85,13 +117,15 @@ export default function Inbox () {
   const { actionableTransactions, recentTransactions } = useTransactionsWithCounterparties()
   const payTransaction = useOffer()
   const declineTransaction = useDecline()
-  const [toggleModal, setToggleModal] = useState(null)
+  const refundTransaction = useRecoverFunds()
+  const [counterpartyNotFound, setCounterpartyNotFound] = useState(true)
+  const [toggleNewTransactionModal, setToggleNewTransactionModal] = useState(null)
```
robbiecarlton: `toggle` is a verb, so in general should only be used for naming a function. (specifically a function that alternates between two states).
I think we want to call this const `isNewTransactionModalVisible`, which is a mouthful, but also says what we mean.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -85,13 +117,15 @@ export default function Inbox () {
...
   const [modalTransaction, setModalTransaction] = useState(null)
 
   const showConfirmationModal = (transaction = {}, action = '') => {
     const modalTransaction = { ...transaction, action }
     if (!isEmpty(transaction) && action !== '') return setModalTransaction(modalTransaction)
-    else return setToggleModal(true)
+    else return setToggleNewTransactionModal(true)
```
robbiecarlton: I'm suspicious of this overloading of `showConfirmationModal`. I'd prefer having a `showNewTransactionModal` function in addition to `showConfirmationModal`.
But I'm not understanding exactly how this part works, so let's talk about this in person, because maybe I'm missing something.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -193,9 +230,17 @@ export function TransactionRow ({ transaction, actionsClickWithTxId, actionsVisi
 
   const isOffer = type === TYPE.offer
   const isRequest = !isOffer
+  const isCanceled = status === STATUS.canceled
+  const isDeclined = status === STATUS.declined
 
   let story
-  if (isActionable) story = isOffer ? ' is offering' : ' is requesting'
+  if (isActionable && !isDeclined) story = isOffer ? ' is offering' : ' is requesting'
```
robbiecarlton, line 245: right now in the style guide we say we're only using one line ifs if they return something. I'm open to re-considering that rule, and this doesn't have to block merging this PR, just mentioning it to keep awareness.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -299,19 +346,19 @@ function PayButton ({ showConfirmationModal, transaction }) {
   </Button>
 }
 
-function RejectButton ({ showConfirmationModal, transaction }) {
-  const action = 'decline'
+function RejectButton ({ showConfirmationModal, transaction, notValid }) {
+  const action = notValid ? 'cancel' : 'decline'
```
robbiecarlton: `notValid={isCanceled || isDeclined}`, so the transaction might be declined, yet we're showing a `Decline` button when `notValid`. This seems like a gap in the logic.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -325,10 +372,22 @@ function NewTransactionModal ({ handleClose, toggleModal }) {
...
 }
 
-export function ConfirmationModal ({ transaction, handleClose, declineTransaction, payTransaction }) {
+export function ConfirmationModal ({ transaction, handleClose, declineTransaction, refundTransaction, payTransaction, setCounterpartyNotFound, counterpartyNotFound }) {
   const { newMessage } = useFlashMessageContext()
-  if (!transaction) return null
-  const { id, counterparty, amount, type, action } = transaction
+  const { id, amount, type, action } = transaction
+  let { counterparty } = transaction
+  if (!counterparty) counterparty = {}
```
robbiecarlton: can replace these lines with `const { counterparty = {} } = transaction`

src/holofuel/pages/Home/Home.js
```diff
@@ -45,7 +51,16 @@ export default function Home () {
...
+  const { newMessage } = useFlashMessageContext()
+  const [errorMessage, setErrorMessage] = useState(null)
+
+  if (errorMessage) {
+    newMessage(errorMessage)
+    setErrorMessage('')
+  }
+
+  const filterActionableTransactionsByStatus = status => holofuelActionableTransactions.filter(actionableTx => actionableTx.status === status)
+  if (!isEmpty(filterActionableTransactionsByStatus('declined')) && errorMessage === null) setErrorMessage(declinedTransactionNotice)
```
robbiecarlton: `newMessage` is a side effect, so you should not make calls to it directly in the render function (on line 58) (no side effects in the render function outside of hooks).
Same with `setErrorMessage` on lines 59 and 63.

you don't actually need state here, what you want is `useEffect`

```
useEffect(() => {
  if (!isEmpty(filterActionableTransactionsByStatus('declined')) && errorMessage === null) {
    newMessage(declinedTransactionNotice)
  } 
}, [filterActionableTransactionsByStatus])
```

So that's how you would achieve this functionality, BUT I'm this seems like bad UX to me, so let's chat and make sure this is how we want to handle this situation, and potentially bring in C and/or Z to think about it.

src/holofuel/pages/CreateRequest/CreateRequest.js
```diff
@@ -0,0 +1,169 @@
+import React, { useState, useEffect } from 'react'
```
robbiecarlton: remove this file

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -174,40 +192,55 @@ export default function CreateOfferRequest ({ history: { push } }) {
...
+  useEffect(() => {
+    setErrorMessage(errorMessage)
+  }, [setErrorMessage, errorMessage])
+
+  useEffect(() => {
+    setCounterpartyNotFound(counterpartyNotFound)
+  }, [setCounterpartyNotFound, counterpartyNotFound])
+
   if (loading) {
+    // TODO: Unsubscribe from Loader to avoid any potential mem leak.
```
robbiecarlton, line 225: Can you help me understand this comment. afaict, if `(!loading)`, `Loader` will be removed from the dom by React in the normal way, no danger of memory leaks.

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -174,40 +192,55 @@ export default function CreateOfferRequest ({ history: { push } }) {
...
   useEffect(() => {
     setCounterpartyNick(nickname)
   }, [setCounterpartyNick, nickname])
 
+  if (!loading && holofuelCounterparty.notFound) {
+    errorMessage = 'This HoloFuel Peer is currently unable to be located in the network. \n Please verify the hash, ensure your HoloFuel Peer is online, and try again after a few minutes.'
+    counterpartyNotFound = true
+  } else if (!loading && !errorMessage && currentCounterpartyValue && currentCounterpartyValue.length === AGENT_ID_LENGTH) {
+    counterpartyNotFound = false
+  } else if (loading) counterpartyNotFound = true
```
robbiecarlton: setting these variables (`errorMessage` and `counterpartyNotFound`) in the render method and then using them in `useEffect` is going to cause subtle bugs. I don't know if it's *technically* "creating side effects in the render method", but it causes the same problems.

Instead, all of this logic should be wrapped in `useEffect`. 

I think it's best if we take a look at this together.

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -62,18 +65,27 @@ export default function CreateOfferRequest ({ history: { push } }) {
...
   useEffect(() => {
     setCounterpartyNick(presentAgentId(counterpartyId))
   }, [counterpartyId])
 
-  const { register, handleSubmit, errors, setValue: setFormValue } = useForm({ validationSchema: FormValidationSchema })
+  const { register, handleSubmit, errors, setValue: setFormValue, getValues } = useForm({ validationSchema: FormValidationSchema })
+  const formValues = getValues()
+
+  if (errorMessage) {
+    newMessage(errorMessage)
```
robbiecarlton: side effect in the render method

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -25,7 +25,10 @@ const AGENT_ID_LENGTH = 63
 const FormValidationSchema = yup.object().shape({
   counterpartyId: yup.string()
     .required()
-    .length(AGENT_ID_LENGTH),
+    // the first four chars === 'HcSc' (ie: the hc prefix)
+    // .substring(0, 5).matches(/(HcSc)/)
```
robbiecarlton: let's remove this commented code.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -137,18 +167,25 @@ function presentTransaction (transaction) {
       if (event.Cheque) return presentCheque({ origin, event, stateDirection, eventTimestamp: timestamp.event, fees: parsedAdjustment.fees, presentBalance: available })
       throw new Error('Completed event did not have a Receipt or Cheque event')
     }
-    case 'rejected': {
-      // We have decided to show this **only** in the inbox page via the recent transactions filter
-      if (event.Request) return presentRequest({ origin, event, stateDirection, eventTimestamp: timestamp.event, fees: parsedAdjustment.fees, status: STATUS.rejected })
-      if (event.Promise) return presentOffer({ origin, event, stateDirection, eventTimestamp: timestamp.event, fees: parsedAdjustment.fees, status: STATUS.rejected })
-      throw new Error('Completed event did not have a Receipt or Cheque event')
+
+    // case 'declined': {
```
robbiecarlton: remove commented code

https://kentcdodds.com/blog/please-dont-commit-commented-out-code

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -105,24 +108,51 @@ const presentCheque = ({ origin, event, stateDirection, eventTimestamp, fees, pr
...
+  if (!declinedTx[2]) return console.log('declinedTx[2] is UNDEFINED : ', declinedTx)
+  const transaction = declinedTx[2].Request ? presentPendingRequest(declinedTx, true) : presentPendingOffer(declinedTx, true)
+  return {
+    ...transaction,
+    status: STATUS.declined
+  }
+}
+
+const presentCanceledTransaction = canceledTx => {
+  if (!canceledTx.event[2].Cancel.entry) return console.log('!canceledTx.event[2].Cancel.entry is UNDEFINED : ', canceledTx)
```
robbiecarlton: should throw

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -195,66 +237,140 @@ const HoloFuelDnaInterface = {
...
-      const noDuplicateIds = _.uniqBy(listOfNonActionableTransactions, 'id')
-      return noDuplicateIds.filter(tx => tx.status === 'pending').sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
+      /* NOTE: Filtering out duplicate IDs should prevent an already completed tranaction from displaying as a pending tranaction if any lag occurs in data update layer.  */
+      const noDuplicateIdsWaitingList = _.uniqBy(listOfNonActionableTransactions, 'id')
+      const listOfDeclinedTransactions = await HoloFuelDnaInterface.transactions.allDeclinedTransactions()
+      // Filter out transactions that share a TX ID with a Declined Transaction (FYI: pending declined txs (on the counterparty side?) will also result from a canceled tx..)
+      const uniqueListWithOutDeclinedOrCanceled = _.differenceBy(noDuplicateIdsWaitingList, listOfDeclinedTransactions, 'id')
+      console.log('noDuplicateIdsWaitingList : ', noDuplicateIdsWaitingList)
+      console.log('ALL WAITING TRANSACTIONS Without Declined or Canceled: ', uniqueListWithOutDeclinedOrCanceled.filter(tx => tx.status === 'pending').sort((a, b) => a.timestamp > b.timestamp ? -1 : 1))
```
robbiecarlton: remove logs

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -195,66 +237,140 @@ const HoloFuelDnaInterface = {
...
+      }
+    },
+    /* NOTE: allNonPending will include Declined and Canceled Transactions:  */
     allNonPending: async () => {
       const { transactions } = await createZomeCall('transactions/list_transactions')()
       const listOfNonActionableTransactions = transactions.map(presentTransaction)
       const noDuplicateIds = _.uniqBy(listOfNonActionableTransactions, 'id')
-      return noDuplicateIds.filter(tx => tx.status !== 'pending').sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
+
+      console.log('ALL NOT_PENDING TRANSACTIONS : ', noDuplicateIds.filter(tx => tx.status !== 'pending').sort((a, b) => a.timestamp < b.timestamp ? -1 : 1))
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -195,66 +237,140 @@ const HoloFuelDnaInterface = {
...
+      }
+    },
+    /* NOTE: This is to allow handling of the other side of the transaction that was declined.  */
+    getPendingDeclined: async (transactionId) => {
+      const declinedResult = await createZomeCall('transactions/list_pending_declined')({ origins: transactionId })
+
+      console.log(' >>>>>>>>>> Pendng Declined transactions RESULT : ', declinedResult)
+
+      const transactionArray = declinedResult.map(presentDeclinedTransaction)
+      console.log('transactionArray :', transactionArray)
```
robbiecarlton: remove logs

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -195,66 +237,140 @@ const HoloFuelDnaInterface = {
...
+    /* NOTE: cancel WAITING TRANSACTION that current agent authored. */
     cancel: async (transactionId) => {
-      const transaction = await HoloFuelDnaInterface.transactions.getPending(transactionId)
-      await createZomeCall('transactions/cancel')({ origin: transactionId })
+      const authoredRequests = await HoloFuelDnaInterface.transactions.allNonActionableByState(transactionId, ['incoming/requested', 'outgoing/approved'])
+      console.log('authoredRequests : ', authoredRequests)
+
+      const transaction = authoredRequests.find(authoredRequest => authoredRequest.id === transactionId)
+
+      console.log(' CANCEL TRANSACTION DETAILS : ', transaction)
```
robbiecarlton: remove logs

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -195,66 +237,140 @@ const HoloFuelDnaInterface = {
...
     }
   },
+  /* NOTE: recover funds from DECLINED PENDING TRANSACTION (ie: Counterparty declined offer) - intended for REFUNDS  */
+  recoverFunds: async (transactionId) => {
+    console.log('INSIDE RECOVER_FUNDS in dna interface, transactionId : ', transactionId)
+
+    const reason = annulTransactionReason
+    const transaction = await HoloFuelDnaInterface.transactions.getPendingDeclined(transactionId)
+
+    console.log('confirm - TRANSACTION TO CANCEL : ', transaction)
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -195,66 +237,140 @@ const HoloFuelDnaInterface = {
...
+    return {
+      ...transaction,
+      id: transactionId,
+      status: STATUS.canceled,
+      canceledReference: canceledProof
+    }
+  },
   requests: {
     create: async (counterpartyId, amount, notes) => {
+      console.log('transaction notes : ', notes)
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -264,13 +380,17 @@ const HoloFuelDnaInterface = {
   },
   offers: {
     create: async (counterpartyId, amount, notes, requestId) => {
-      const origin = await createZomeCall('transactions/promise')(pickBy(i => i, { to: counterpartyId, amount: amount.toString(), deadline: MOCK_DEADLINE, notes, requestId }))
+      console.log('transaction notes : ', notes)
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -279,7 +399,10 @@ const HoloFuelDnaInterface = {
     },
 
     accept: async (transactionId) => {
+      console.log('ACCEPT transactionId : ', transactionId)
+
       const transaction = await HoloFuelDnaInterface.transactions.getPending(transactionId)
+      console.log('ACCEPT transaction : ', transaction)
```
robbiecarlton: remove logs

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -297,13 +420,13 @@ const HoloFuelDnaInterface = {
     acceptMany: async (transactionIdArray) => {
       const result = await createZomeCall('transactions/receive_payments_pending')({ promises: transactionIdArray })
       const transactionArray = Object.entries(result).map(presentAcceptedPayment)
-      return transactionArray.sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
+      return transactionArray.sort((a, b) => a.timestamp < b.timestamp ? -1 : 1).reverse()
```
robbiecarlton: rather than calling reverse here, you can just change the logic in the `sort` function. Either switching the `-1` and `1` or replacing the `<` with `>` will do it.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -297,13 +420,13 @@ const HoloFuelDnaInterface = {
...
     },
 
     acceptAll: async () => {
       const result = await createZomeCall('transactions/receive_payments_pending')({})
       const transactionArray = Object.entries(result).map(presentAcceptedPayment)
-      return transactionArray.sort((a, b) => a.timestamp < b.timestamp ? -1 : 1)
+      return transactionArray.sort((a, b) => a.timestamp < b.timestamp ? -1 : 1).reverse()
```
robbiecarlton: see above for changing the sorting logic

src/graphql-server/resolvers.js
```diff
@@ -99,9 +99,20 @@ export const resolvers = {
...
+    holofuelCancel: (_, { transactionId }) => {
+      console.log('!! INSIDE CANCEL ENDPOINT IN RESOLVERS !!')
+      return HoloFuelDnaInterface.transactions.cancel(transactionId)
+    },
 
-    holofuelCancel: (_, { transactionId }) => HoloFuelDnaInterface.transactions.cancel(transactionId),
+    holofuelRecoverFunds: (_, { transactionId }) => {
+      console.log('!! INSIDE RECOVR FUNDS ENDPOINT IN RESOLVERS !!')
+      return HoloFuelDnaInterface.transactions.recoverFunds(transactionId)
+    },
```
robbiecarlton: remove logs and make these one line definitions, like `holofuelAcceptOffer` above

src/graphql-server/schema.graphql
```diff
@@ -66,6 +73,8 @@ type Transaction {
   presentBalance: Float
   notes: String
   happName: String
+  declinedReference: String
+  canceledReference: String
```
robbiecarlton: How are we using these values in the front end? We only need to put stuff in graphql types that the frontend needs access to.

src/graphql-server/schema.graphql
```diff
@@ -125,17 +134,18 @@ type Query {
...
-type Mutation {
+type Mutation { 
   registerUser(name: String, avatarUrl: String): Me
   registerHostingUser: HhaUser
   enableHapp(appId: String): Happ
   disableHapp(appId: String): Happ
   updateHostPricing(units: String, pricePerUnit: String): HostPricing
-  holofuelRequest(counterpartyId: String, amount: Float): Transaction
-  holofuelOffer(counterpartyId: String, amount: Float, requestId:ID): Transaction
+  holofuelRequest(counterpartyId: String, amount: Float, notes: String): Transaction
```
robbiecarlton, line 136: good catch 😂

src/graphql/HolofuelCompletedTransactionsQuery.gql
```diff
@@ -1,5 +1,5 @@
-query HolofuelCompletedTransactionsQuery {
-  holofuelCompletedTransactions {
+query HolofuelCompletedTransactionsQuery ($transactionId: String)  {
+  holofuelCompletedTransactions(filters: $transactionId)  {
```
robbiecarlton: this filter variable is not plugged in in the schema and resolvers so won't do anything.

src/holochainClient.js
```diff
@@ -2,7 +2,7 @@ import { connect as hcWebClientConnect } from '@holochain/hc-web-client'
 import { get } from 'lodash/fp'
 import mockCallZome from 'mock-dnas/mockCallZome'
 
-const developmentMockDnaConnection = true // this is the value MOCK_DNA_CONNECTION will have in the dev server
+const developmentMockDnaConnection = false // this is the value MOCK_DNA_CONNECTION will have in the dev server
```
robbiecarlton: should be true on develop

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -17,11 +18,14 @@ export function PrimaryLayout ({
...
 }) {
-  const { data: { holofuelActionableTransactions: transactions = [] } = {} } = useQuery(HolofuelActionableTransactionsQuery, { fetchPolicy: 'network-only' })
+  const { data: { holofuelActionableTransactions: actionableTransactions = [] } = {} } = useQuery(HolofuelActionableTransactionsQuery, { fetchPolicy: 'network-only' })
   const { loading: holofuelUserLoading, data: { holofuelUser = {} } = {} } = useQuery(HolofuelUserQuery)
   const { data: { holofuelLedger: { balance: holofuelBalance } = { balance: 0 } } = {} } = useQuery(HolofuelLedgerQuery, { fetchPolicy: 'network-only' })
 
-  const inboxCount = transactions.length
+  const inboxCount = actionableTransactions.filter(actionableTx => (actionableTx.status !== STATUS.canceled) && (actionableTx.type !== TYPE.request)).length
+  console.log('>>>>>>>> actionableTransactions.length : ', actionableTransactions.length)
+  console.log('>>>>>>>> inboxCount : ', inboxCount)
```
robbiecarlton: remove logs

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -17,11 +18,14 @@ export function PrimaryLayout ({
   children,
   headerProps = {}
 }) {
-  const { data: { holofuelActionableTransactions: transactions = [] } = {} } = useQuery(HolofuelActionableTransactionsQuery, { fetchPolicy: 'network-only' })
+  const { data: { holofuelActionableTransactions: actionableTransactions = [] } = {} } = useQuery(HolofuelActionableTransactionsQuery, { fetchPolicy: 'network-only' })
   const { loading: holofuelUserLoading, data: { holofuelUser = {} } = {} } = useQuery(HolofuelUserQuery)
   const { data: { holofuelLedger: { balance: holofuelBalance } = { balance: 0 } } = {} } = useQuery(HolofuelLedgerQuery, { fetchPolicy: 'network-only' })
 
-  const inboxCount = transactions.length
+  const inboxCount = actionableTransactions.filter(actionableTx => (actionableTx.status !== STATUS.canceled) && (actionableTx.type !== TYPE.request)).length
```
robbiecarlton: I made similar comments on `Inbox`, and looking here it's even clearer to me. We should be handling this with different graphql queries. There should at least be two different queries, one that returns *actually* actionable transactions (ie not cancelled), and a different one that returns *all* actionable transactions.

src/holofuel/pages/CreateOffer/CreateOffer.test.js
```diff
@@ -0,0 +1,261 @@
+import React from 'react'
```
robbiecarlton: remove this file.

robbiecarlton (changes requested):
Let's talk on video about some of these changes. Some of them are a little involved.

package.json
JettTech: Totally agree. :100:

src/holochainClient.js
JettTech: So, I've implemented Global Vars to take care of this issue because I so easily default to testing it live and then pushing the updates without updating this variable each time. :thinking:  ...But now the script will handle this update for us... so neither of us have to worry about defaulting to the wrong value!! :rocket:

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
JettTech: Let's look at this together. There is actually a React Warning in the console that states such.  I'll show you when we pair. :eyes:

src/data-interfaces/__mocks__/HoloFuelInterface.js
JettTech: It's just stating that the transaction type could be an offer or a request in a case of synchronous transactions, which is *not* the case for asynchronous transactions.  (During an asynchronous offer, the counterparty is 'pre-approved' for an incoming offers from offer author, resulting in any offer automatically being accepted and soon after, affecting each party's resulting balance. Thus in the asynchronous transactions, *only* requests can be canceled.)

JettTech: We discussed this and agreed the comment isn't needed and we will handle the testing edge case for requests if/when it surfaces.

src/holofuel/pages/Inbox/Inbox.js
JettTech: Agreed, reason demo'd live. ;)

src/holofuel/pages/Inbox/Inbox.js
JettTech: Just to document from our last convo, this is actually showing a 'Cancel' button whenever the transaction has already been declined. (There is one edge case wherein an offer has been declined and we will need to cancel this declined offer to restore the full balance back from the escrow.)  The isValid is also used for styling the declined and canceled transactions.

Conclusion: During our last conversation, we decided to pass in `isCanceled` and `isDeclined` separately to handle each case and then ref both for styles.

src/holofuel/pages/Inbox/Inbox.js
JettTech: **We'll look back at this once the rest updated. (This is the approach I prefer as well, but for some reason the counterparty value wasn't being set properly thereafter.)

src/holofuel/pages/Inbox/Inbox.js
JettTech: Totally agree about the desire to abstract away the business logic into the dna interface. 

*To document what I shared in our last conversation,  I chose to go with this approach to avoid more endpoint calls that are subject to the interim fetch:network-only method we're using currently to overcoming any polling/caching issues.  Essentially, too many calls, too frequently are occurring with each 'getall'-type call at the moment.  Let's look at moving this as an endpoint on the interface once we refactor with subscriptions or different caching approaches.*

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: Agree and decided on this approach (only for now) to avoid adding additional heavy calls to the page.  Refer to previous comment in Inbox.

src/graphql/HolofuelCompletedTransactionsQuery.gql
JettTech: Ah!  This is actually residue from when I did have it plugged in.  (I decided not to use this filter for reasons relating to advanced enums within javascript vs Rust. This is one of the items we need to review/include in our api revision during a code refactoring after launch.)

Nonetheless, looks like I deleted it everywhere, but here.  Good catch! :eye: :eye:

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: Agreed, and left comments as to why this is implemented in this case (for now).  See comment above. :+1:

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
JettTech: I do think we should implement a prefix check, but I'll remove the comment here and just make a PT chore for this. :+1:

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Hmm... Strange!! ...I actually updated this to exactly what you stated here (and it's implemented for `allCompleted`, `allActionable`, and `allWaiting`...but it seems that was overwritten in a merge from 1170 to 4770).  Anyway, I agree it's preferable and will updatdeit accordingly. :dart:

src/holofuel/pages/Home/Home.js
JettTech: Yes, this was a placeholder message for a larger conversation.  NB: Had the conversation with Z and Rob yesterday, this message will become a modal with a bulk action associated with it.  I will leave this placeholder for now.. as this will be the foundation upon which we can use for the updates.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -195,57 +230,110 @@ const HoloFuelDnaInterface = {
...
-        status: STATUS.cancelled
+        status: STATUS.canceled
       }
     }
   },
+  /* NOTE: recover funds from DECLINED PENDING TRANSACTION (ie: Counterparty declined offer) - intended for REFUNDS  */
+  recoverFunds: async (transactionId) => {
+    const reason = annulTransactionReason
+    const transaction = await HoloFuelDnaInterface.transactions.getPendingDeclined(transactionId)
+    const canceledProof = await createZomeCall('transactions/cancel')({ entry: transaction, reason })
```
robbiecarlton, line 329: there are two endpoints in the DNA called `transactions/cancel` and `transactions/cancel_transactions` and they do different things?

src/graphql-server/schema.graphql
```diff
@@ -35,6 +35,12 @@ type HostPricing {
 type HolofuelUser {
   id: ID
   nickname: String
+  notFound: Boolean
+}
+
+type ListOrigins {
```
robbiecarlton: This type is not used anywhere, so should be removed.  

(if we were keeping it we'd want to name it a noun not a verb, as it's a type. Probably `OriginList`)

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -62,12 +63,14 @@ export default function CreateOfferRequest ({ history: { push } }) {
...
   const { data: { holofuelHistoryCounterparties: agents } = {} } = useQuery(HolofuelHistoryCounterpartiesQuery)
-
   const createOffer = useOfferMutation()
   const createRequest = useRequestMutation()
 
+  const { newMessage } = useFlashMessageContext()
+
   const [counterpartyId, setCounterpartyId] = useState('')
   const [counterpartyNick, setCounterpartyNick] = useState('')
+  const [isCounterpartyFound, setCounterpartyFound] = useState(null)
```
robbiecarlton: This should be initialized false not null, `useState(false)`

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -174,40 +181,59 @@ export default function CreateOfferRequest ({ history: { push } }) {
...
+  useEffect(() => {
+    if (!loading) {
+      if (notFound) {
+        setCounterpartyFound(false)
+        if (!hasDisplayedNotFoundMessage) {
+          newMessage('This HoloFuel Peer is currently unable to be located in the network. \n Please verify the hash, ensure your HoloFuel Peer is online, and try again after a few minutes.')
+          setHasDisplayedNotFoundMessage(true)
+        }
+      } else {
+        setCounterpartyFound(true)
```
robbiecarlton, line 215: you need to reset `hasDisplayedNotFoundMessage` here, and in the other else clause.
`setHasDisplayedNotFoundMessage(false)`

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -174,40 +181,59 @@ export default function CreateOfferRequest ({ history: { push } }) {
...
+      }
+    } else {
+      setCounterpartyFound(false)
```
robbiecarlton, line 219: add `setHasDisplayedNotFoundMessage(false)`

src/holofuel/pages/Home/Home.js
```diff
@@ -45,7 +51,15 @@ export default function Home () {
...
   const goToOfferRequest = () => history.push(OFFER_REQUEST_PATH)
 
-  const { data: { holofuelLedger: { balance: holofuelBalance } = { balance: 0 } } = {} } = useQuery(HolofuelLedgerQuery, { fetchPolicy: 'network-only' })
+  const { newMessage } = useFlashMessageContext()
+
+  const filterActionableTransactionsByStatus = useCallback(status => holofuelActionableTransactions.filter(actionableTx => actionableTx.status === status), [holofuelActionableTransactions])
+
+  useEffect(() => {
+    if (!isEmpty(filterActionableTransactionsByStatus('declined'))) {
+      newMessage(declinedTransactionNotice)
```
robbiecarlton, line 60: I'm not a fan of this UX, but let's call this the MVP and bring it the UXers for implementation in a new ticket.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -85,15 +110,19 @@ export default function Inbox () {
...
+  const refundTransaction = useRefund()
+  const [counterpartyNotFound, setCounterpartyNotFound] = useState(true)
+  const [isNewTransactionModalVisible, setIsNewTransactionModalVisible] = useState(null)
   const [modalTransaction, setModalTransaction] = useState(null)
 
   const showConfirmationModal = (transaction = {}, action = '') => {
     const modalTransaction = { ...transaction, action }
     if (!isEmpty(transaction) && action !== '') return setModalTransaction(modalTransaction)
-    else return setToggleModal(true)
+    return null
```
robbiecarlton: We don't care about the return value of `showConfirmationModal`, we only call it for it's side effect, so we should not return anything from it. 
Line 120 should be 
```
if (!isEmpty(transaction) && action !== '') setModalTransaction(modalTransaction)
```
and you can remove the `return null` line altogether

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -85,15 +110,19 @@ export default function Inbox () {
   const { actionableTransactions, recentTransactions } = useTransactionsWithCounterparties()
   const payTransaction = useOffer()
   const declineTransaction = useDecline()
-  const [toggleModal, setToggleModal] = useState(null)
+  const refundTransaction = useRefund()
+  const [counterpartyNotFound, setCounterpartyNotFound] = useState(true)
+  const [isNewTransactionModalVisible, setIsNewTransactionModalVisible] = useState(null)
```
robbiecarlton: this should be initialized to false `useState(false)`.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -299,19 +345,19 @@ function PayButton ({ showConfirmationModal, transaction }) {
...
+    <p>{isDeclined ? 'Cancel' : 'Decline'}</p>
   </Button>
 }
 
-function NewTransactionModal ({ handleClose, toggleModal }) {
+function NewTransactionModal ({ handleClose, isNewTransactionModalVisible }) {
   return <Modal
     contentLabel={'Create a new transaction.'}
-    isOpen={!!toggleModal}
+    isOpen={!!isNewTransactionModalVisible}
```
robbiecarlton: this is close to nitpicking, but because `isNewTransactionModalVisible` is a bool, `!!` is redundant.

`isNewTransactionModalVisible === !!isNewTransactionModalVisible` for all values of `isNewTransactionModalVisible`

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -325,10 +371,21 @@ function NewTransactionModal ({ handleClose, toggleModal }) {
...
   const { newMessage } = useFlashMessageContext()
-  if (!transaction) return null
-  const { id, counterparty, amount, type, action } = transaction
+  const { id, amount, type, action } = transaction
+  const { counterparty = {} } = transaction
+  const { holofuelCounterparty, loading } = useCounterparty(counterparty.id)
+
+  useEffect(() => {
+    if (!transaction) return null
+    else if (holofuelCounterparty || loading) {
```
robbiecarlton: the `loading` condition here doesn't do anything, because it will never be the case that both `loading` and `holofuelCounterparty.notFound` are true at the same time.

robbiecarlton (changes requested):
A few more simple changes. But we're very close!

src/holofuel/pages/Home/Home.js
JettTech: Yes, agreed; this is exactly what this is. :100:   

I actually spoke to Zoe and Rob about this on Friday, and an interim UX decision was made, which is now reflected in Figma (basically a modal - instead of a flash  message - that triggers a bulk action to refund any/all outstanding transactions). The new PT ticket for this is: https://www.pivotaltracker.com/story/show/170394140. 

We can definitely PL this more in stand-up tomorrow though, as you suggested. :+1:

src/graphql-server/schema.graphql
JettTech: This was meant to be removed when it was from the interface and resolvers. Good eye.

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Yes, they do. The prior is for a proactive cancel, allowing the initiating agent to cancel an offer/request prior to the counterparty taking action.   The later is in response to a decline action by the counterparty (ie: basically for restoring any the money from a cancelled offer back to the initiating agent's balance).

src/holofuel/pages/Inbox/Inbox.js
JettTech: Good point. :+1:  I was using the previous modal code as boilerplate, and just didn't update after it we pegged the isOpen value to a boolean.  This is a good edit, even though it's small.

src/holofuel/pages/Inbox/Inbox.js
JettTech: Good catch!

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
JettTech: I thought I added these state resets.. not sure when/how it was deleted. :thinking: Nonetheless, good call.

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
JettTech: Same as above...

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -374,13 +373,13 @@ function NewTransactionModal ({ handleClose, isNewTransactionModalVisible }) {
 export function ConfirmationModal ({ transaction, handleClose, declineTransaction, refundTransaction, payTransaction, setCounterpartyNotFound, counterpartyNotFound }) {
   const { newMessage } = useFlashMessageContext()
   const { id, amount, type, action } = transaction
-  const { counterparty = {} } = transaction
-  const { holofuelCounterparty, loading } = useCounterparty(counterparty.id)
+  const { counterparty = {}, notFound } = transaction
```
robbiecarlton: `notFound` shouldn't come from transaction here, but from `holofuelCounterparty` on the line below. We should also have a test that would catch this mistake.

robbiecarlton (changes requested):

src/holofuel/pages/Inbox/Inbox.js
JettTech: I don't think this should block the merge given all that is dependent.  I will make a simple test for this, but otherwise am in favor of making a Chore in PT to update/upgrade some of our testing for measures like this.

zo-el (approved):

robbiecarlton (approved):

PR #135, opened 2019-12-10
Updated the holo-nixpkgs to the latest release

zo-el:
- hc Version 0.0.34-alpha4
- removed agent2 from config.nix because you can only have one agent on a conductor according to current settings.

> Note: We should talk about setting up two conductors for this repo. And plans to use `hc run` cmd

### DIFF
--- a/config.nix
+++ b/config.nix
@@ -9,13 +9,6 @@
     keystore_file = "agent.key";
     public_address = "HcScIio76M8cbDqhrsZKp37W8YZ4y8X9v9epzbyhRzqpjeb5sigdC4puN6b5koi";
   }
-  # Currently we can only run one agent on a conductor
-  #{
-  #  id = "hp-admin-agent-2";
-  #  name = "HP Admin 2";
-  #  keystore_file = "agent-2.key";
-  #  public_address = "HcScjSPMA9g46x6nosEih9thqwvesx53udZEkqMfqcwnbw9sZKPSyIFNvtfxrqi";
-  #}
   ];
 
 }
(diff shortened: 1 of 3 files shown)


config.nix
```diff
@@ -9,11 +9,13 @@
...
-    keystore_file = "agent-2.key";
-    public_address = "HcScjSPMA9g46x6nosEih9thqwvesx53udZEkqMfqcwnbw9sZKPSyIFNvtfxrqi";
-  }];
+  # Currently we can only run one agent on a conductor
+  #{
+  #  id = "hp-admin-agent-2";
+  #  name = "HP Admin 2";
+  #  keystore_file = "agent-2.key";
+  #  public_address = "HcScjSPMA9g46x6nosEih9thqwvesx53udZEkqMfqcwnbw9sZKPSyIFNvtfxrqi";
+  #}
```
lukateras: https://kentcdodds.com/blog/please-dont-commit-commented-out-code
https://stackoverflow.com/questions/3664288

lukateras (commented):
LGTM otherwise.

config.nix
robbiecarlton: <3

robbiecarlton (changes requested):
Just remove the commented code and it's good to merge.

config.nix
zo-el: Reason: This was left in because when we spin up two conductors soon (This will be done soon. Hopefully this week). We would need this.

**But I am happy to remove the commented code out.**

robbiecarlton (approved):

PR #136, opened 2019-12-11
4620 - HoloFuel - Collapse send and receive screens

robbiecarlton:


JettTech:
Exciting functionality updates and LOC reduction!!  A smile bursts on my face as I see the UX simplicity take shape and our codebase less redundant!! :rainbow: 

Also, I have a question for you.  When reviewing this PR, I noticed that the hi-fi design implementation isn't implemented.  Are we prioritizing the functional with lo-fi designs and iterating over later with the hi-fi?  I'd just like to ensure I have the current a/c in mind when reviewing...

PR #138, opened 2019-12-15
fixed display more than one inbox message

zo-el:


JettTech (approved):
This does the trick.  Nice scouting and solution-ing!! :raised_hands:

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -238,7 +238,7 @@ const HoloFuelDnaInterface = {
     },
     allActionable: async () => {
       const { requests, promises, declined, canceled } = await createZomeCall('transactions/list_pending')()
-      const actionableTransactions = requests.map(presentPendingRequest).concat(promises.map(presentPendingOffer)).concat(declined.map(presentDeclinedTransaction)).concat(canceled.map(presentCanceledTransaction))
+      const actionableTransactions = requests.map(r => presentPendingRequest(r)).concat(promises.map(p => presentPendingOffer(p))).concat(declined.map(presentDeclinedTransaction)).concat(canceled.map(presentCanceledTransaction))
```
robbiecarlton, line 241: @zo-el - This code appears to do the same thing as before the change to me, will you help me understand how this fixes the bug?

JettTech: @robbiecarlton - It's because `presentPendingRequest` and `presentPendingOffer` now takes a second param to handle declined/canceled requests and offers; however, since the map fn includes a default second param of index, when handling the pending Requests/Offers, we need to explicit pass only the first param.

PR #144, opened 2019-12-23
Quick fixes

zo-el:


JettTech (approved):
Thanks for adding the placeholder loader. :+1:

PR #147, opened 2019-12-28
8934(agent can no longer request TestFuel from themselves successfully), 0932(HoloFuel form handling - Block negative numbers in Send/Receive), & 7556 (Set Default TF loading value - instead of displaying 0)

JettTech:


zo-el (approved):

src/components/FlashMessage/FlashMessage.test.js
```diff
@@ -36,7 +37,7 @@ it('should render a closable message', () => {
 it('should render a message which closes itself after the given time', async () => {
   const message = 'this is the message'
   const time = 10
-  const { queryByText, getByText } = render(
+  const { queryByText, getByText } = await renderAndWait(
```
robbiecarlton: what was the reason for this change?

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -37,7 +43,7 @@ export function PrimaryLayout ({
       agent={holofuelUser}
       agentLoading={holofuelUserLoading}
       inboxCount={inboxCount}
-      holofuelBalance={holofuelBalance}
+      holofuelBalance={ledgerNullState || holofuelBalance}
```
robbiecarlton: I think this could be simpler and easier to read. How about just passing `ledgerLoading` as a separate prop to `SideMenu`, and then in `SideMenu` you can branch on that, rather than testing if `holofuelBalance` is `NaN`.

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -211,6 +216,12 @@ export function RenderNickname ({ agentId, setCounterpartyNick, setCounterpartyF
           newMessage('This HoloFuel Peer is currently unable to be located in the network. \n Please verify the hash, ensure your HoloFuel Peer is online, and try again after a few minutes.')
           setHasDisplayedNotFoundMessage(true)
         }
+      } else if (id === whoami.id) {
+        setCounterpartyFound(false)
+        if (!hasDisplayedNotFoundMessage) {
+          newMessage('You cannot send yourself TestFuel.')
+          setHasDisplayedNotFoundMessage(true)
+        }
```
robbiecarlton: I think we should do this check in the parent component. We already know our own agentID, so when the counterparty field changes, we can just compare that to our agentId, and show a message then (and disable the button on that condition as well as on `counterpartyNotFound`).

lmk if that's not clear and we can look at it together

src/holofuel/pages/Home/Home.js
```diff
@@ -38,9 +38,14 @@ function useTransactionsWithCounterparties () {
   }
 }
 
+const DisplayBalance = ({ ledgerLoading, holofuelBalance }) => {
+  if (ledgerLoading) return <>-- TF</>
+  else return <>{presentHolofuelAmount(holofuelBalance)} TF</>
+}
+
```
robbiecarlton, line 45: exactly like this for the `SideMenu`.

robbiecarlton (changes requested):

src/components/FlashMessage/FlashMessage.test.js
JettTech: When updating the `createOfferRequest` tests, which trigger a flash message, I received an error with this flash message test and used the `await renderAndWait` to ensure the flash message wasn't prompted until all the mock queries were complete.  Since then, I updated the `createOfferRequest` tests with a missing mock Query, which also resolved .  Consequently, this test will work with or without the `renderAndWait` now.

robbiecarlton: great, will you put it back to regular `render`?

robbiecarlton: if this works with `render` we should use that.

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -62,19 +63,32 @@ const modePrepositions = {
...
 
   const [counterpartyId, setCounterpartyId] = useState('')
   const [counterpartyNick, setCounterpartyNick] = useState('')
   const [isCounterpartyFound, setCounterpartyFound] = useState(false)
+  const [isFormValid, setIsFormValid] = useState(false)
 
   useEffect(() => {
     setCounterpartyNick(presentAgentId(counterpartyId))
-  }, [counterpartyId])
+    if (counterpartyId.length === AGENT_ID_LENGTH && counterpartyId === whoami.id) {
```
robbiecarlton: the length check here is redundant. `counterpartyId === whoami.id` implies `counterpartyId.length === AGENT_ID_LENGTH`

robbiecarlton: otherwise, good change 👍

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -87,10 +101,16 @@ export default function CreateOfferRequest ({ history: { push } }) {
   const [total, setTotal] = useState(0)
 
   const onAmountChange = amount => {
-    if (isNaN(amount)) return
-    const newFee = Number(amount) * FEE_PERCENTAGE
-    setFee(newFee)
-    setTotal(Number(amount) + newFee)
+    if (isNaN(amount)) return null
```
robbiecarlton, line 107: don't need to explicitly return `null` here and it's slightly confusing. Makes it seem like we care about the return value when actually we're just exiting early.

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -186,19 +208,19 @@ export default function CreateOfferRequest ({ history: { push } }) {
         agents={agents}
         selectedAgentId={counterpartyId}
         selectAgent={selectAgent} />
-      <Button type='submit' dataTestId='submit-button' wide variant='secondary' styleName='send-button' disabled={counterpartyId.length === AGENT_ID_LENGTH ? !isCounterpartyFound : true}>Send</Button>
+      <Button type='submit' dataTestId='submit-button' wide variant='secondary' styleName='send-button' disabled={counterpartyId.length !== AGENT_ID_LENGTH || !isCounterpartyFound || !isFormValid}>Send</Button>
```
robbiecarlton: This is fine as it is, but I think it would be neater to have a `useEffect` that sets the value of `isFormValid` and takes into account all of the factors (the ones here and the ones already determining `isFormValid`)

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -219,7 +241,7 @@ export function RenderNickname ({ agentId, setCounterpartyNick, setCounterpartyF
       setCounterpartyFound(false)
       setHasDisplayedNotFoundMessage(false)
     }
-  }, [setCounterpartyFound, setHasDisplayedNotFoundMessage, hasDisplayedNotFoundMessage, loading, notFound, newMessage])
+  }, [setCounterpartyFound, setHasDisplayedNotFoundMessage, hasDisplayedNotFoundMessage, loading, notFound, newMessage, whoami, id])
```
robbiecarlton: looks like this `useEffect` doesn't depend on `whoami`

robbiecarlton (changes requested):

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
JettTech: This is purely to abide by the linting rules. I agree that it's unnecessary otherwise.

robbiecarlton: there's no linting rule that specifies you have to return a value.

`if (isNaN(amount)) return` should make the linter happy

src/components/FlashMessage/FlashMessage.test.js
```diff
@@ -37,7 +36,7 @@ it('should render a closable message', () => {
 it('should render a message which closes itself after the given time', async () => {
   const message = 'this is the message'
   const time = 10
-  const { queryByText, getByText } = await renderAndWait(
+  const { queryByText, getByText } = render(
```
robbiecarlton: 😄😄😄

robbiecarlton (approved):

PR #148, opened 2019-12-30
2019 12 29 tweaks for deployment

peeech:
Tweaks of the code for making HP Admin UI work with HPOS

### DIFF
--- a/src/holochainClient.js
+++ b/src/holochainClient.js
@@ -23,7 +23,7 @@ export const HOLOCHAIN_LOGGING = true && process.env.NODE_ENV !== 'test'
 // Parse window.location to retrieve holoPort's HC public key (3rd level subdomain in URL)
 const getHcPubkey = () => {
   return ((process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test')
-    ? '3llrdmlase6xwo9drzs6qpze40hgaucyf7g8xpjze6dz32s957'
+    ? '5m5srup6m3b2iilrsqmxu6ydp8p8cr0rdbh4wamupk3s4sxqr5'
     : window.location.hostname.split('.')[0])
 }
 
@@ -43,6 +43,7 @@ export const getHpAdminKeypair = async (email = undefined, password = undefined)
   try {
     const hcKey = getHcPubkey()
     if (!hcKey || !email || !password) return null
+
     const HpAdminKeypair = await importHpAdminKeypairClass()
     HpAdminKeypairInstance = new HpAdminKeypair(hcKey, email, password)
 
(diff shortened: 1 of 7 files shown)


src/holochainClient.js
```diff
@@ -108,7 +108,7 @@ let holochainClient
 async function initAndGetHolochainClient () {
   if (holochainClient) return holochainClient
   try {
-    let url = process.env.NODE_ENV === 'production' ? undefined : process.env.REACT_APP_DNA_INTERFACE_URL
+    let url = process.env.NODE_ENV === 'production' ? '/api/v1/ws/' : process.env.REACT_APP_DNA_INTERFACE_URL
```
robbiecarlton: I was surprised by this change, so just want to double check. this is the base url that's passed to `hc-web-client`. My understanding is that we want to leave that undefined in production as the conductor handles that. Are you sure you want it to be this path?

peeech: `connect` ([source](https://github.com/holochain/hc-web-client/blob/master/src/index.ts#L29)) needs to be passed a url at which conductor is listening. Our conductor happens to be running on the HPOS, so we need to provide a URL to HPOS. That's where nginx on HPOS comes into play via [those settings](https://github.com/Holo-Host/holo-nixpkgs/blob/develop/profiles/logical/hpos/default.nix#L114) and redirects calls to the conductor instance running on HPOS.

robbiecarlton: ok, got it. In that case it's good to merge as far as I'm concerned.

robbiecarlton (approved):

src/holochainClient.js
```diff
@@ -108,11 +109,11 @@ let holochainClient
 async function initAndGetHolochainClient () {
   if (holochainClient) return holochainClient
   try {
-    let url = process.env.NODE_ENV === 'production' ? undefined : process.env.REACT_APP_DNA_INTERFACE_URL
-    // Construct url with query param X-Holo-Admin-Signature = signature
+    let url = process.env.NODE_ENV === 'production' ? '/api/v1/ws/' : process.env.REACT_APP_DNA_INTERFACE_URL
```
JettTech: Good update, @peeech. :+1:

src/holochainClient.js
```diff
@@ -23,7 +23,7 @@ export const HOLOCHAIN_LOGGING = true && process.env.NODE_ENV !== 'test'
 // Parse window.location to retrieve holoPort's HC public key (3rd level subdomain in URL)
 const getHcPubkey = () => {
   return ((process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test')
-    ? '3llrdmlase6xwo9drzs6qpze40hgaucyf7g8xpjze6dz32s957'
+    ? '5m5srup6m3b2iilrsqmxu6ydp8p8cr0rdbh4wamupk3s4sxqr5'
```
JettTech, line 26: How is this pub key being generated?  Can we make it less static?

PR #150, opened 2019-12-30
8626 - HP Admin Routing changes

robbiecarlton:
Don't merge until the parent branch has been merged

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -97,3 +97,7 @@ function formatDateTime (isoDate) { // eslint-disable-line no-unused-vars
     // Throw Error, iso-timedate cannot be parsed into valid format
   } else throw new Error('Iso timedate is unable to be parsed.', isoDate)
 }
+
+export function getCommunityUrl () {
+  return window.location.protocol + '//' + window.location.host + '/community/'
+}
--- a/__mocks__/@holo-host/hp-admin-keypair.js
+++ b/__mocks__/@holo-host/hp-admin-keypair.js
@@ -1,0 +1,8 @@
+
+exports.HpAdminKeypair = class {
+  constructor (hcKey, email, password) { } // eslint-disable-line no-useless-constructor
+
+  sign (payload) {
+    return 'signature'
+  }
+}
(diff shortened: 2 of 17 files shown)


robbiecarlton:
@zo-el - here are the routing changes you asked for. I ended up merging in some changes PJ had made in another branch, which should be merged soon, so it's not ready to merge, but it's ready for review.

zo-el (approved):
LGTM :+1:

src/utils/index.js
```diff
@@ -97,3 +97,7 @@ function formatDateTime (isoDate) { // eslint-disable-line no-unused-vars
     // Throw Error, iso-timedate cannot be parsed into valid format
   } else throw new Error('Iso timedate is unable to be parsed.', isoDate)
 }
+
+export function getCommunityUrl () {
+  return window.location.protocol + '//' + window.location.host + '/community/'
+}
```
JettTech, line 103: Has this approach been tested yet?

JettTech (approved):
I'd like to talk a bit more about testing the Community Link.  Otherwise, glad we went with the `/admin/` approach for now.

lukateras (changes requested):
I'd like to request hash-based routing to be considered instead. It will overall work much better with our routing and won't require special case handling like this approach does. All it would take to implement is switching from `BaseRouter` to [`HashRouter`](https://reacttraining.com/react-router/web/api/HashRouter) in [`src/root.js`](https://github.com/Holo-Host/hp-admin/blob/db41a48292112e525060219a7c9617c5d2738ab3/src/root.js#L4).

src/utils/index.js
robbiecarlton: I've tested it in that it produces the expected url in the browser. So if community app is on `{host}/community/` then this will work. But I doubt it's been tested in HPOS with community app running.

peeech: I will be able to test it on HPOS later this week.

robbiecarlton:
@yegortimoshenko - if it's not blocking MVP release, let's come back to that once we've shipped. I'm open to it, but would prefer not to make more changes than necessary right now.

PR #151, opened 2019-12-30
updated the preseed data using two conductors

zo-el:


scripts/initialize-conductor-data.js
```diff
@@ -43,25 +42,26 @@ async function populateHoloFuelData () {
 
   // 4.) Scenario: A request from agent1 to agent2, which agent 2 has paid :
```
zo-el, line 44: This Scenario is giving some issues and is not consistent. 
Due to time restrictions, I have just commented it out.

src/utils/const.js
```diff
@@ -0,0 +1,4 @@
+export const agent1 = {
+  "id" : "HcScIio76M8cbDqhrsZKp37W8YZ4y8X9v9epzbyhRzqpjeb5sigdC4puN6b5koi",
```
robbiecarlton: this is brittle compared to reading it out of the conductor directly. Is it not possible to read it from the conductor anymore?

scripts/initialize-conductor-data.js
zo-el: This scenario does not work because of some networking issues. The data is not propagated on time.

src/utils/const.js
zo-el: Done

default.nix
```diff
@@ -232,7 +232,7 @@ in
       {
         id = "websocket-interface";
         driver = {
-          port = 3400;
+          port = 3401;
```
JettTech, line 235: :tada:

scripts/get-agent.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs')
```
JettTech, line 1: This file is almost identical to `scripts/prepareHoloFuelData/agent-config.js`.  Can we reference this file or somehow merge these, to avoid duplicate code?

scripts/create-zome-call.js
```diff
@@ -1,12 +1,36 @@
-const { findInstanceForAgent } = require('../src/utils/integration-testing/conductorConfig')
-const initAndGetHolochainClient = require('./holochain-client.js')
+const { connect: hcWebClientConnect } = require('@holochain/hc-web-client')
```
JettTech: These updates create a file almost identical to `scripts/holochain-client`.  Can we reference this file instead of the extra LOC?

JettTech (changes requested):

robbiecarlton (approved):

JettTech (approved):

PR #152, opened 2020-01-02
(Also PAUSED) 9561 - Implement pagination on Transaction History

robbiecarlton:
After discussions with @pjkundert about how HoloFuel DNA pagination works, I've decided to go with local pagination for MVP. We'll revisit this after launch.

### DIFF
--- a/src/data-interfaces/HoloFuelDnaInterface.js
+++ b/src/data-interfaces/HoloFuelDnaInterface.js
@@ -240,7 +240,7 @@ const HoloFuelDnaInterface = {
       return actionableTransactions.sort((a, b) => a.timestamp > b.timestamp ? -1 : 1)
     },
     allWaiting: async () => {
-      const { transactions } = await createZomeCall('transactions/list_transactions')()
+      const { transactions } = await createZomeCall('transactions/list_transactions')()      
       const listOfNonActionableTransactions = transactions.map(presentTransaction)
       /* NOTE: Filtering out duplicate IDs should prevent an already completed tranaction from displaying as a pending tranaction if any lag occurs in data update layer.  */
       const noDuplicateIdsWaitingList = _.uniqBy(listOfNonActionableTransactions, 'id')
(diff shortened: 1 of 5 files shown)


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -252,6 +252,7 @@ const HoloFuelDnaInterface = {
     },
     allDeclinedTransactions: async () => {
       const declinedResult = await createZomeCall('transactions/list_pending_declined')()
+      console.log('declinedResult', declinedResult)
```
JettTech: Delete log. :wink:

src/mock-dnas/holofuel.js
```diff
@@ -421,6 +485,7 @@ const holofuel = {
     ledger_state: () => transactionList.ledger,
     list_transactions: () => transactionList,
     list_pending: listPending,
+    list_pending_declined: [],
```
JettTech, line 488: :+1:

src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -123,6 +128,8 @@ export default function TransactionsHistory () {
       </React.Fragment>)}
     </div>}
 
+    {hasMoreTransactions && <Button onClick={showMoreTransactions}>Load More</Button>}
```
JettTech, line 131: Let's add the conditional that omits this button component whenever there are no transactions.
ie: `{!noVisibleTransactions && hasMoreTransactions && <Button onClick={showMoreTransactions}>Load More</Button>}`

JettTech (changes requested):
I have requested a few touch ups, but looks good for an interim pagination approach.  I'd also like to talk over this new approach and understand why Perry said to avoid the use of the hf dna pagination logic.

src/holofuel/pages/TransactionHistory/TransactionHistory.js
robbiecarlton: awkwardly, it's possible for there to be more transactions to load, even while there are `noVisibleTransactions`. (eg: there are no withdrawals in the current transaction list, but there are some later.) This is a slightly broken UX, and we can make it smarter, but I want to get a first iteration in before we ship.

PR #154, opened 2020-01-02
6271 - HoloFuel: all urls redirect to home page

robbiecarlton:
@JettTech - ok this is ready for review

### DIFF
--- a/src/contexts/useAuthContext.js
+++ b/src/contexts/useAuthContext.js
@@ -3,7 +3,7 @@ import React, { createContext, useContext, useState } from 'react'
 export const AuthContext = createContext()
 
 export function AuthProvider ({ children }) {
-  const [isAuthed, setIsAuthed] = useState(false)
+  const [isAuthed, setIsAuthed] = useState(process.env.NODE_ENV === 'development')
 
   return <AuthContext.Provider value={{ isAuthed, setIsAuthed }}>
     {children}
--- a/src/holofuel/HFRouter.js
+++ b/src/holofuel/HFRouter.js
@@ -1,26 +1,17 @@
 import React from 'react'
-import { Route, useRouteMatch } from 'react-router-dom'
+import { Route, Redirect } from 'react-router-dom'
 
 import Home from 'holofuel/pages/Home'
 import Inbox from 'holofuel/pages/Inbox'
 import TransactionHistory from 'holofuel/pages/TransactionHistory'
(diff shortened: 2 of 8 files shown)


src/holofuel/HFRouter.test.js
```diff
@@ -56,7 +56,7 @@ describe('HFRouter', () => {
     testLinks(<MemoryRouter initialEntries={['/holofuel']}><HFRouter /></MemoryRouter>)
   })
 
-  it('works from /holofuel/', () => {
+  it.skip('works from /holofuel/', () => {
```
robbiecarlton: I'm skipping this test because I reached the end of my rope. This change works fine in the app and I just haven't figured out why the test won't pass.

robbiecarlton: ok I figured it out

src/contexts/useAuthContext.js
```diff
@@ -3,7 +3,7 @@ import React, { createContext, useContext, useState } from 'react'
 export const AuthContext = createContext()
 
 export function AuthProvider ({ children }) {
-  const [isAuthed, setIsAuthed] = useState(false)
+  const [isAuthed, setIsAuthed] = useState(process.env.NODE_ENV === 'development')
```
JettTech, line 6: Good update. :+1:

src/holofuel/HFRouter.js
```diff
@@ -1,26 +1,17 @@
...
-    <Route path={makePath('/(|home)')} exact component={Home} />
-    <Route path={makePath('/inbox')} exact component={Inbox} />
-    <Route path={makePath('/history')} component={TransactionHistory} />
-    <Route path={makePath('/tos')} exact component={Tos} />
-    <Route path={makePath('/offer-request')} component={CreateOfferRequest} />
+    <Route path='/holofuel' exact component={() => <Redirect to='/holofuel/' />} />
+    <Route path='(/|/holofuel/)(|home)' exact component={Home} />
+    <Route path='(/|/holofuel/)inbox' exact component={Inbox} />
+    <Route path='(/|/holofuel/)history' component={TransactionHistory} />
+    <Route path='(/|/holofuel/)offer-request' component={CreateOfferRequest} />
```
JettTech, line 15: What is the reason we don't have 'exact' path for these routes?

JettTech (approved):
Overall, looks good.  I have one question, I'd like to follow-up on, but I don't consider it blocking.

PR #156, opened 2020-01-03
4024 - Implement HoloFuel MVP v.0.1.3 UI design - Transaction History

robbiecarlton:


src/holofuel/components/SideMenu/SideMenu.js
```diff
@@ -57,7 +57,7 @@ export default function SideMenu ({
             </Link>
           </li>
           <li>
-            <Link to='/makemework' styleName='nav-link'>
+            <Link to='/admin/' styleName='nav-link'>
```
JettTech, line 60: This leads to an empty page, which will cause the app to break when served via HoloScape and will not connect properly in Holochain contexts.  

For now, I think we can leave this as is and update it whenever we do our build compatibility between holochain and holo contexts.  One thing I would like to discuss is using hash routing (ie: `<Link to='/#admin/' styleName='nav-link'>`), which might help avoid collision between contexts...

src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -107,24 +102,24 @@ export default function TransactionsHistory () {
...
+          ledgerLoading={ledgerLoading} />
       </div>
+
+      <FilterButtons filter={filter} setFilter={setFilter} />
     </div>
-    <FilterButtons filter={filter} setFilter={setFilter} />
 
     {noVisibleTransactions && <div styleName='transactions-empty'>
       You have no {transactionTypeName}.
     </div>}
```
JettTech: Let's update the styles for the null state page as well.

JettTech (changes requested):
The updates here look good; however, I noticed that the null state pages are not implemented.  Can you please include those, and ping for another review?  Almost there! :)

JettTech (approved):
LGTM

src/holofuel/components/SideMenu/SideMenu.js
robbiecarlton: This link just shouldn't appear outside of `hp-admin` context, so that's part of a broader issue we need to figure out.

PR #157, opened 2020-01-04
Updated the holochainClient to make one WebSocket Connection

zo-el:


JettTech (approved):
LGTM. Thanks for updating this. :tada:  

Also, happy to leave the error catch to 10 tries/call for now, but let's talk about this and see if there is an optimal number.

src/holochainClient.js
```diff
@@ -131,14 +133,28 @@ async function initAndGetHolochainClient () {
...
 }
+async function initAndGetHolochainClient () {
+  let counter = 0
+  while (hcFlag) {
+    counter++
+    await wait(100)
+    if (counter === 10) {
+      hcFlag = false
+    }
+  }
```
robbiecarlton: isn't this equivalent to
```
if (hcFlag) {
  await wait(1000)
  hcFlag = false
}
```
?

src/holochainClient.js
```diff
@@ -111,9 +112,10 @@ export function conductorInstanceIdbyDnaAlias (instanceId) {
 }
 
 let holochainClient
+let hcFlag = false
 
-async function initAndGetHolochainClient () {
-  if (holochainClient) return holochainClient
+async function init () {
```
robbiecarlton: on the fence about this, but I think I'd prefer this be called `initHolochainClient`.

It makes the specific difference between the two functions clearer.

src/holochainClient.js
```diff
@@ -111,9 +112,10 @@ export function conductorInstanceIdbyDnaAlias (instanceId) {
 }
 
 let holochainClient
+let hcFlag = false
```
robbiecarlton: can we call this something more descriptive?

`isInitiatingHcConnection` or similar

package.json
```diff
@@ -20,7 +20,7 @@
     "test:lint": "standard",
     "test:unit": "node scripts/test.js --config=./jest.unit.config.js",
     "test": "npm run test:lint && npm run test:unit",
-    "test:ci": "CI=1 npm run test",
+    "test:ci": "CI=1 npm run test:unit",
```
robbiecarlton, line 24: 😆 I feel like we make this change all the time and somehow it never sticks

robbiecarlton (changes requested):

src/holochainClient.js
robbiecarlton: oh, I get it.

Going to think about if there's a more readable way to do this.

src/holochainClient.js
```diff
@@ -131,14 +133,28 @@ async function initAndGetHolochainClient () {
...
     if (HOLOCHAIN_LOGGING) {
       console.log('😞 Holochain client connection failed -- ', error.toString())
     }
+    isInitiatingHcConnection = false
     throw (error)
   }
 }
+async function initAndGetHolochainClient () {
+  let counter = 0
+  while (isInitiatingHcConnection) {
```
robbiecarlton, line 154: can we add a comment calling out that this variable is being changed in a different function running in parallel?

robbiecarlton (changes requested):
One more change request, then it's good to merge 👍

robbiecarlton (approved):

PR #161, opened 2020-01-09
0647 - HoloFuel - Decimal value rounding in numpad & Request Fees

robbiecarlton:


src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -139,12 +141,12 @@ export default function CreateOfferRequest ({ history: { push } }) {
       <h4 styleName='amount-label'>
         {title}
       </h4>
-      <div styleName='amount'>
+      <div styleName='amount' onClick={() => setNumpadVisible(true)}>
```
JettTech, line 144: Nice addition.  Did you met with C to determine whether this should be accompanied with a icon to signal this functionality?

JettTech (changes requested):
Almost there! :+1:  

At [line #22 in AmountInput.js](https://github.com/Holo-Host/hp-admin/blob/29577d00d02e9554d6fdf05518b26428d28336a3/src/holofuel/pages/CreateOfferRequest/AmountInput/AmountInput.js#L22), we need to handle the condition wherein we only have one digit in the InputValue string, and a user selects the `removeDigit` function.  (Otherwise, when processing the new InputValue inside the `presentHolofuelString`, the split method will fail and result in an error.) 

Example Implementation:
```
const isValidinputValueLength = inputValue.length > 1
const removeDigit = () => isValidinputValueLength ? setInputValue(String(inputValue).slice(0, -1)) :  setInputValue(String(0))
```

JettTech:
Also, I noticed a small CSS bug.  It looks like the `.amount-input-amount ` class inside AmountInput.module.css needs to explicitly set its width to 100%.  Otherwise, the input box stretches well beyond the width of the screen and causes alignment and odd interaction/behavior while in a mobile view.

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
robbiecarlton: no this is more like an easter egg. It just seemed like the right thing to do. There could definitely be some kind of design element to point people to this feature.

src/holofuel/pages/CreateOfferRequest/AmountInput/AmountInput.js
```diff
@@ -8,7 +8,7 @@ export default function AmountInput ({ amount, setAmount, chooseSend, chooseRequ
   const [inputValue, setInputValueRaw] = useState(String(amount))
 
   const setInputValue = value => {
-    const cleanValue = value.replace(/[^0-9.]/g, '') || 0 // strips non numerical characters
+    const cleanValue = value.replace(/[^0-9.]/g, '') || '0' // strips non numerical characters
```
JettTech, line 11: :clap:

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.test.js
```diff
@@ -446,4 +446,22 @@ describe('AmountInput', () => {
     fireEvent.click(getByText('Send'))
     expect(getByText(`${presentHolofuelAmount('1.034')} TF`)).toBeInTheDocument()
   })
+
```
JettTech, line 449: Good job on adding the extra test! :+1:

JettTech (approved):
LGTM!  Nice catch on the string type. :shipit:

PR #168, opened 2020-01-15
7847 - HoloFuel CleanUp Items

robbiecarlton:


src/holofuel/HFRouter.js
```diff
@@ -1,17 +1,17 @@
...
 import CreateOfferRequest from 'holofuel/pages/CreateOfferRequest'
 
 export default function HFRouter () {
-  return <>
+  return <Switch>
+    <Route path='/holofuel/(|home)' exact component={Home} />
+    <Route path='/holofuel/inbox' exact component={Inbox} />
+    <Route path='/holofuel/history' exact component={TransactionHistory} />
+    <Route path='/holofuel/offer-request' exact component={CreateOfferRequest} />
     <Route path='/holofuel' exact component={() => <Redirect to='/holofuel/' />} />
```
JettTech, line 15: Let's add a route for the Holochain case, to make sure we still support HF testing in / use with Holoscape.  We could do something like the following:

`{process.env.REACT_APP_HOLOFUEL_APP === 'true' && <Route path='/' exact component={() => <Redirect to='/holofuel/' />} />}`

src/holofuel/HFRouter.test.js
```diff
@@ -46,13 +46,9 @@ const testLinks = ui => {
   expect(getByText('TransactionHistory')).toBeInTheDocument()
 }
 
-// we test three cases here, urls of the form '/', '/holofuel' and '/holofuel/'
+// we test two cases here, urls of the form '/holofuel' and '/holofuel/'
 
 describe('HFRouter', () => {
-  it('works with root url', () => {
```
JettTech, line 52: We should probably still keep this case, for serving HF in Holochain contexts, like Holoscape.

src/graphql-server/schema.graphql
```diff
@@ -68,6 +68,7 @@ type Transaction {
   presentBalance: Float
   notes: String
   happName: String
+  isPayingARequest: Boolean
```
JettTech, line 71: :+1:

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -152,7 +154,8 @@ function presentPendingOffer (transaction, declined = false) {
   const eventTimestamp = event[1]
   const counterpartyId = declined ? event[2].Promise.tx.to : provenance[0]
   const { amount, notes, fee } = event[2].Promise.tx
-  return presentOffer({ origin, event: event[2], stateDirection, status, type, eventTimestamp, counterpartyId, amount, notes, fees: fee })
+  const isPayingARequest = !!event[2].Promise.request
```
JettTech, line 161: Nice. Glad this solution worked as expected. :+1:

src/holofuel/pages/Inbox/Inbox.test.js
```diff
@@ -35,12 +34,15 @@ const actionableTransactions = pendingList.requests.concat(pendingList.promises)
...
       counterparty: item.event[2].Promise.tx.from,
-      timestamp: item.event[1]
+      timestamp: item.event[1],
+      isPayingARequest: !!item.event[2].Promise.request
     }
   } else {
     throw new Error('unrecognized transaction type', item.toString())
   }
 })
+  .sort((a, b) => a.timestamp > b.timestamp ? -1 : 1)
```
JettTech: Nice change.  I think I reflected this same change in my most recent PR too! :muscle: :100:

JettTech (changes requested):
Overall, looks good!  Glad to see the `isPayingARequest` update.  

There is one change we should make to avoid the app opening onto an empty page for the holochain  (& thereby holoscape) environments, but otherwise, I think these changes are straightforward. :+1:

src/holofuel/HFRouter.js
robbiecarlton: Good call.
I forget about the Holoscape testing case.

package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.2.1",
   "private": true,
   "scripts": {
-    "start-mock": "rm -rf ./node_modules/.cache/babel-loader && REACT_APP_MOCK_DNA_CONNECTION='true' node scripts/start.js",
-    "start-live": "rm -rf ./node_modules/.cache/babel-loader && REACT_APP_MOCK_DNA_CONNECTION='false' node scripts/start.js",
+    "start": "npm run start:mock",
+    "start:mock": "rm -rf ./node_modules/.cache/babel-loader && REACT_APP_MOCK_DNA_CONNECTION='true' node scripts/start.js",
+    "start:live": "rm -rf ./node_modules/.cache/babel-loader && REACT_APP_MOCK_DNA_CONNECTION='false' node scripts/start.js",
```
robbiecarlton, line 8: @JettTech NB this small update. Just standardizing the naming.

src/holofuel/HFRouter.js
robbiecarlton: because of the way the routers are nested, I could actually add it without the condition.

package.json
```diff
@@ -3,8 +3,9 @@
...
     "start:holofuel-agent-1": "PORT=3100 REACT_APP_DNA_INTERFACE_URL=ws://localhost:3400 npm run start:holofuel-live",
     "start:holofuel-agent-2": "PORT=3101 REACT_APP_DNA_INTERFACE_URL=ws://localhost:3401 npm run start:holofuel-live",
     "start:holofuel-mock": "REACT_APP_HOLOFUEL_APP=true npm run start-mock",
```
JettTech: Due to the change in the names, lines 11 & 12 need ot have the script updated from `start-mock` and `start-live` to `start:mock` and `start:live`

JettTech (changes requested):
Ok.  I requested a simple update due to the script name changes.  Otherwise, your code looks good and unit tests pass.

JettTech:
@robbiecarlton - FYI: In order to manually test this branch against live data, I updated the nix-config and created new agents.  During my attempts to test in a two-agent scenario, I encountered extreme networking issues, to the point that I could only test this branch in a single agent environment.  0.0.42-alpha4 is *noticeably slower*. 

I don't think a two-agent testing scenario is necessary for this ticket and therefore think this is good to ship as soon as the script names are updated. Nonetheless, I think it is important to note, and I will make a new ticket in our backlog to represent this need.

package.json
robbiecarlton: good catch

PR #172, opened 2020-01-22
updated the hc v0.0.42-alpha5

zo-el:


robbiecarlton (approved):

JettTech (approved):

JettTech:
@zo-el - I approved this PR, because we need this new version for upgraded hdk functionality.  

However noticed that the login-flow test is failing.  Is there something specific about this version of HC that would cause this test to fail on CI here, but not in previous versions? This particular test was always a bit erratic locally (which is a TODO), but in my experience, it never failed a CI.

PR #173, opened 2020-01-24
0056 - Chore: Fix useEffect / useState pattern where we check state to see if an effect has run

robbiecarlton:


src/contexts/useFlashMessageContext.js
```diff
@@ -1,14 +1,16 @@
...
+import React, { createContext, useContext, useState, useCallback } from 'react'
 
 export const FlashMessageContext = createContext()
 
 export function FlashMessageProvider ({ children }) {
   const [message, setMessage] = useState('')
   const [time, setTime] = useState()
-  const newMessage = (message, time) => {
+
+  const newMessage = useCallback((message, time) => {
```
robbiecarlton, line 9: this is the change that fixed the issue (alongside the matching change in HoloFuel). Everything else in this PR is just removing our workarounds and testing.

JettTech (approved):
LTGM. When you have a moment, I'd to discuss the thoughts behind the use of useCallback here.

PR #175, opened 2020-01-27
2093 - Update hp-admin to work in HPOS

robbiecarlton:


default.nix
```diff
@@ -1,4 +1,4 @@
-{ pkgs ? import ./pkgs.nix {}, shell ? false }:
+{ pkgs ? import ./nixpkgs.nix {} }:
```
robbiecarlton: @zo-el the original PR did't remove this `shell ? false` expression, that was something I did when resolving merge conflicts. Wasn't sure the right thing to do, maybe we want to keep that expression?

shell.nix
```diff
@@ -1,4 +1,4 @@
-{ pkgs ? import ./pkgs.nix {}, shell ? false  }:
+{ pkgs ? import ./nixpkgs.nix {} }:
```
robbiecarlton: @zo-el - see comment on `default.nix` above

default.nix
zo-el: We do need that. It will give you access to `sim2h_server` cli

zo-el:
```suggestion
{ pkgs ? import ./nixpkgs.nix {}, shell ? false }:
```

shell.nix
zo-el:
```suggestion
{ pkgs ? import ./nixpkgs.nix {}, shell ? false }:
```

robbiecarlton:
@zo-el ok, I incorporated the changes you suggested. LMK if this is good to merge

zo-el (approved):
Can we squash and merge this branch?

PR #177, opened 2020-02-03
0029 hpos connection user feedback

JettTech:


src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -1,5 +1,6 @@
-import React, { useContext, useState } from 'react'
+import React, { useContext, useState, useEffect } from 'react'
 import { object } from 'prop-types'
+// import { isEmpty } from 'lodash/fp'
```
robbiecarlton: remove comment

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -19,7 +22,23 @@ export function PrimaryLayout ({
...
+  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: 30000, onError })
+  const { newMessage } = useFlashMessageContext()
+
+  useEffect(() => {
+    if (!isConnected) {
+      newMessage('Your Holoport is currently unreachable.', 0)
+    } else {
+      newMessage('', 0)
+    }
+  }, [isConnected, setIsConnected, newMessage])
```
robbiecarlton, line 44: It's so nice we can write this now and don't have to add a whole extra layer of `hasMessageBeenDisplayed` logic :)

src/root.js
```diff
@@ -7,6 +7,8 @@ import apolloClient from 'apolloClient'
 import ReactModal from 'react-modal'
 import HFRouter from './holofuel/HFRouter'
 import ScreenWidthContext from 'contexts/screenWidth'
+import { ConnectionProvider } from 'contexts/useConnectionContext'
+// import { WhoamiProvider } from 'contexts/useWhoamiContext'
```
robbiecarlton: remove comment

src/contexts/useWhoamiContext.js
```diff
@@ -0,0 +1,15 @@
+import React, { createContext, useContext, useState } from 'react'
```
robbiecarlton: afaict we're not using this so please remove this file.

src/apolloClient.js
```diff
@@ -1,12 +1,39 @@
...
 import apolloLogger from 'apollo-link-logger'
 import { SchemaLink } from 'apollo-link-schema'
+import { onError } from 'apollo-link-error'
 import { InMemoryCache } from 'apollo-cache-inmemory'
 import schema from 'graphql-server'
 
+const errorLink = onError(({ graphQLErrors, networkError, response }) => {
+  if (networkError) {
+    console.log(`[Network error]: ${networkError}`)
+    response.errors.hposConnection = false
```
robbiecarlton: This variable name is confusing.

`response.errors.hposConnection` reads to me as "is there an error in the hposConnection?" not "do we have a working hposConnection?"

Ideally we would put it somewhere else in the response object, and rename it to something like:

`response.isHposConnectionActive`

Would have to dig into the docs to see if that would be possible. But even just renaming it

`response.errors.isHposConnectionActive`

would make it much clearer.

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -19,7 +22,23 @@ export function PrimaryLayout ({
   showSideMenu = true,
   showAlphaFlag = true
 }) {
-  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery)
+  const { setIsConnected, isConnected } = useConnectionContext()
+
+  const onError = ({ graphQLErrors }) => {
+    const hposConnection = graphQLErrors.hposConnection
```
robbiecarlton: nbd but you could also write this as 
`const { hposConnection } = graphQLErrors`

src/contexts/useConnectionContext.js
```diff
@@ -0,0 +1,15 @@
+import React, { createContext, useContext, useState } from 'react'
```
robbiecarlton, line 1: this context is a great addition 👍 👍 👍

robbiecarlton (changes requested):
I might be missing something, but I can't see anywhere where `isConnected` is set to true, except in the case of an authentication error. If that's the case, then how does this work during normal operation. Seems like the login form will just always be disabled.

src/contexts/useWhoamiContext.js
JettTech: Done.  This was spill-over; I ended up extracting this work into a different branch, but didn't double-check the cleanup.

JettTech:
@robbiecarlton  - So a user should technically *always* encounter the Authentication Error when first arriving at the login page, as the call to the settings query is done at page load, and thus, prior to the user ever logging in and having their auth keys to successfully wrap the calls to HPOS. 

Therefore, you are correct in your reading and it is intentional that `isConnected` is only set to true when there is an authentication error.

zo-el (changes requested):
`isConnected` should also be set to `true` when you get a successful response from the query.
Currently, you are only updating `isConnected` when there is an error (`onError`)
You have to update it `onComplete` as well (This is when you do not face an error while making the call).

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -19,7 +21,24 @@ export function PrimaryLayout ({
...
 }) {
-  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery)
+  const { setIsConnected, isConnected } = useConnectionContext()
+
+  const onError = ({ graphQLErrors }) => {
+    const { isHposConnectionActive } = graphQLErrors
+    setIsConnected(isHposConnectionActive)
+  }
+
+  const { error, data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: 10000, onError, notifyOnNetworkStatusChange: true, ssr: false })
```
robbiecarlton: what's the purpose of the `ssr` flag here?

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -19,7 +21,24 @@ export function PrimaryLayout ({
...
+  const { newMessage } = useFlashMessageContext()
+
+  useEffect(() => {
+    if (!isConnected) {
+      console.log('error.include(is Error) : ', error)
```
robbiecarlton: remove log

robbiecarlton (changes requested):
+1 to @zo-el s comment. Currently once the user is logged in, `isConnected` will be false for the whole session, meaning we can't use `useConnectionContext` for anything else.

src/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: This ensures that the login page still appears when the server is down and the connection isn't made. (Since we included the `notifyOnNetworkStatusChange` option, this catches any change in connection to the server and stops the rendering of the page whenever it is unable to connect, so we will need this new ssr flag in place to accommodate the first page-load.)

src/data-interfaces/HposInterface.js
```diff
@@ -22,7 +22,7 @@ export function hposCall ({ method = 'get', path, apiVersion = 'v1', headers: us
     return mockCallHpos(method, apiVersion, path)
   } else {
     return async params => {
-      const fullPath = ((process.env.NODE_ENV === 'production') ? (window.location.protocol + '//' + window.location.hostname) : process.env.REACT_APP_HPOS_URL) + '/api/' + apiVersion + '/' + path
+      const fullPath = ((process.env.NODE_ENV === 'production') ? (window.location.protocol + '//' + window.location.hostname) : 'https://2ah5s30i2dldd5wh30iq5c1jd0hmmz6hin0c2g59c5id9znawv.holohost.net') + '/api/' + apiVersion + '/' + path //  process.env.REACT_APP_HPOS_URL
```
robbiecarlton: bring back env variable

robbiecarlton (changes requested):
One change, plus the merge conflict, which you could actually fix by just changing the version number in package.json to "0.10.0" and then running `npm install`, rather than actually merging and resolving the conflicts.

JettTech:
@zo-el @robbiecarlton - I had originally thought that when the connection is lost, we would require the page would redirect to the login (which I thought we'd also handle in 7292).  

I now understand the we are not progressing with that approach, and so yes we will need (and now have) a case, wherein the connection is set to true when we already have keys generated and the polling call to settings is successful. - Thank you for bringing this to my attention. :+1:

src/data-interfaces/HposInterface.js
JettTech: Yes! :+1: Good catch!

robbiecarlton (approved):

robbiecarlton (approved):
LGTM!

zo-el (approved):
Looks good now :+1:

PR #179, opened 2020-02-05
7292 - As a user when I refresh a hApp page it takes me back to the login screen.

robbiecarlton:


JettTech (approved):
Merge it! :shipit:

PR #182, opened 2020-02-10
4360 hc compatible holofuel

JettTech:
- wraps routes in logical wrapper to discern between hpos and holochain env - to be compatible with holochain env
- updates ws url setting in holochainClient to also work with hc environment


src/holochainClient.js
```diff
@@ -113,15 +113,20 @@ let holochainClient
...
   try {
-    let url = process.env.NODE_ENV === 'production' ? ('wss://' + window.location.hostname + '/api/v1/ws/') : process.env.REACT_APP_DNA_INTERFACE_URL
-    // Construct url with query param X-Hpos-Admin-Signature = signature
-    const urlObj = new URL(url)
-    const params = new URLSearchParams(urlObj.search.slice(1))
-    params.append('X-Hpos-Admin-Signature', await signPayload('get', urlObj.pathname))
-    params.sort()
-    urlObj.search = params.toString()
-    url = urlObj.toString()
+    if (process.env.REACT_APP_HOLOFUEL_APP === 'true') {
```
zo-el: Can we rename this `REACT_APP_HOLOFUEL_APP` env var?
Because its use is unclear.

zo-el (changes requested):

src/holochainClient.js
JettTech: :+1: Good point, this name is a bit outdated now that HoloFuel is bundled within hp-admin when served through the HPOS, and is really only a pure "Holofuel" app whenever run within a Holochain environment. 

Considering the current distinction, I purpose renaming it to : `REACT_APP_HOLOCHAIN_APP`.  Thoughts?

zo-el: Its is more like `holofuel in holochain mode`.

zo-el (approved):

src/holochainClient.js
JettTech: Ok decided on `REACT_APP_HOLOFUEL_AS_HOLOCHAIN_APP` and pushed that up in the last update.

src/holofuel/HFRouter.js
```diff
@@ -1,19 +1,26 @@
...
-import { Switch, Redirect } from 'react-router-dom'
-
+import { Route, Switch, Redirect } from 'react-router-dom'
 import Home from 'holofuel/pages/Home'
 import Inbox from 'holofuel/pages/Inbox'
 import TransactionHistory from 'holofuel/pages/TransactionHistory'
 import CreateOfferRequest from 'holofuel/pages/CreateOfferRequest'
 
+function HFRoute (props) {
+  if (process.env.REACT_APP_HOLOFUEL_AS_HOLOCHAIN_APP === 'true') {
```
robbiecarlton: This should use `REACT_APP_HOLOFUEL_APP` (see overall review comment)

src/holochainClient.js
```diff
@@ -113,15 +113,20 @@ let holochainClient
...
   try {
-    let url = process.env.NODE_ENV === 'production' ? ('wss://' + window.location.hostname + '/api/v1/ws/') : process.env.REACT_APP_DNA_INTERFACE_URL
-    // Construct url with query param X-Hpos-Admin-Signature = signature
-    const urlObj = new URL(url)
-    const params = new URLSearchParams(urlObj.search.slice(1))
-    params.append('X-Hpos-Admin-Signature', await signPayload('get', urlObj.pathname))
-    params.sort()
-    urlObj.search = params.toString()
-    url = urlObj.toString()
+    if (process.env.REACT_APP_HOLOFUEL_AS_HOLOCHAIN_APP === 'true') {
```
robbiecarlton: This should use `REACT_APP_RAW_HOLOCHAIN` (see overall review comment)

src/holofuel/components/SideMenu/SideMenu.js
```diff
@@ -56,7 +56,7 @@ export default function SideMenu ({
               History
             </Link>
           </li>
-          {process.env.REACT_APP_HOLOFUEL_APP !== 'true' && <li>
+          {process.env.REACT_APP_HOLOFUEL_AS_HOLOCHAIN_APP !== 'true' && <li>
```
robbiecarlton: This should use `REACT_APP_HOLOFUEL_APP` (see overall review comment)

src/root.js
```diff
@@ -15,7 +15,7 @@ import { FlashMessageProvider as HFFlashMessageProvider } from 'holofuel/context
 import HPAdminRouter from './HPAdminRouter'
 
 export function App () {
-  if (process.env.REACT_APP_HOLOFUEL_APP) {
+  if (process.env.REACT_APP_HOLOFUEL_AS_HOLOCHAIN_APP === 'true') {
```
robbiecarlton: This should use `REACT_APP_HOLOFUEL_APP` (see overall review comment)

robbiecarlton (changes requested):
This is definitely something we need to do. However, I think this implementation collapses two distinct questions that we want to keep separate:

1) Are we running JUST HoloFuel, or all of hp-admin?

2) Are we running inside of HPOS, or as a raw holochain app.

So I think we want two different env variables:

`REACT_APP_HOLOFUEL_APP` (which we already have)
and
`REACT_APP_RAW_HOLOCHAIN` (or something. I can't find a name I love for this var, if you have something better use that)

robbiecarlton (approved):

PR #187, opened 2020-02-17
Update holo nixpkgs

zo-el:


JettTech (approved):
This last one works and we'll need it pushed into the rc to work for testing on the hpos. :pray: :ship: :shipit:

PR #188, opened 2020-02-18
0444 feedback color change & 9430 Inbox Amount Color Bug

JettTech:


src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -188,14 +193,9 @@ export default function Inbox ({ history: { push } }) {
...
 
-  const disableActionedTransaction = transactionId => {
-    if (hasTransactionBeenActioned.idList) {
-      return hasTransactionBeenActioned.idList.find(txid => txid === transactionId)
-    } else return false
-  }
-  const increaseAction = () => includes(DIRECTION.incoming, hasTransactionBeenActioned)
-  const decreseAction = () => includes(DIRECTION.outgoing, hasTransactionBeenActioned)
-  const statusQuoAction = () => includes('status-quo', hasTransactionBeenActioned)
+  const findTransactionById = (transactionId, arrayOfTransactions) => arrayOfTransactions.find(txid => txid === transactionId)
```
robbiecarlton: you can define this in the TransactionRow component as that's the only place that uses it. Also, instead of `arrayOfTransactions`, can we call the argument `transactions`? the plural tells us it's an array.

src/holofuel/pages/TransactionHistory/TransactionHistory.js
```diff
@@ -240,9 +250,19 @@ export function ConfirmCancellationModal ({ transaction, handleClose, cancelTran
...
       newMessage(`${capitalize(type)} succesfully cancelled.`, 5000)
+      setLastActionSuccess(id)
+      clear(5000)
     }).catch(() => {
       newMessage('Sorry, something went wrong', 5000)
+      setLastActionError(id)
+      clear(5000)
     })
+
+    const clear = timeout => setTimeout(() => {
```
robbiecarlton: move this definition above where it's being called.

robbiecarlton (changes requested):
let's get rid of `showConfirmationModal`, and change `setModalTransaction` to 

`setConfirmationModalProperties`

```
confirmationModalProperties = {
  isDisplayed: Boolean,
  transaction: Transaction,
  action: String ( one of 'pay', 'accept', 'cancel', 'decline'),
  onConfirm: function
}
```

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -147,7 +147,28 @@ export default function Inbox ({ history: { push } }) {
   const refundAllDeclinedTransactions = useRefundAllDeclinedTransactions()
   const [counterpartyNotFound, setCounterpartyNotFound] = useState(true)
   const [isDeclinedTransactionModalVisible, setIsDeclinedTransactionModalVisible] = useState(false)
-  const [modalTransaction, setModalTransaction] = useState(null)
+  const [confirmationModalProperties, setConfirmationModalProperties] = useState({
+    shouldDisplay: false,
+    transactions: [],
+    action: '',
+    hasConfirmed: false
```
robbiecarlton: this should be stored in a const

`const defaultConfirmationModalProperties = { ... }`

and used here and in the definition of `resetDefaultModalTransaction`

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -485,9 +483,12 @@ export function DeclinedTransactionModal ({ handleClose, isDeclinedTransactionMo
   </Modal>
 }
 
-export function ConfirmationModal ({ transaction, handleClose, declineTransaction, refundTransaction, payTransaction, acceptOffer, setCounterpartyNotFound, counterpartyNotFound, setHasTransactionBeenActioned }) {
+export function ConfirmationModal ({ confirmationModalProperties, setNewModalTransactionProperties, clearHighlightedTransaction, declineTransaction, refundTransaction, payTransaction, acceptOffer, setCounterpartyNotFound, counterpartyNotFound }) {
   const { newMessage } = useFlashMessageContext()
-  const { id, amount, type, action, notes } = transaction
+  const { transactions, action, shouldDisplay } = confirmationModalProperties
+  const transaction = transactions[0] || {}
```
robbiecarlton: why are we passing in a list of transactions here if we're only ever using the first one?

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -147,7 +147,28 @@ export default function Inbox ({ history: { push } }) {
...
+      hasConfirmed: false
+    })
+  }
+
+  const setNewModalTransactionProperties = (newProperties = {}) => {
+    if (isEmpty(newProperties)) resetDefaultModalTransaction()
+    else setConfirmationModalProperties(newProperties)
+  }
+
+  const clearHighlightedTransaction = timeout => setTimeout(() => resetDefaultModalTransaction(), timeout)
```
robbiecarlton: this name seems inaccurate to me. Doesn't this close the modal?

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -147,7 +147,28 @@ export default function Inbox ({ history: { push } }) {
...
   const [isDeclinedTransactionModalVisible, setIsDeclinedTransactionModalVisible] = useState(false)
-  const [modalTransaction, setModalTransaction] = useState(null)
+  const [confirmationModalProperties, setConfirmationModalProperties] = useState({
+    shouldDisplay: false,
+    transactions: [],
+    action: '',
+    hasConfirmed: false
+  })
+
+  const resetDefaultModalTransaction = () => {
```
robbiecarlton: I think we can call this `hideConfirmationModal` and it's easier to read and not really lying

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -147,7 +147,28 @@ export default function Inbox ({ history: { push } }) {
...
+      transactions: [],
+      action: '',
+      hasConfirmed: false
+    })
+  }
+
+  const setNewModalTransactionProperties = (newProperties = {}) => {
+    if (isEmpty(newProperties)) resetDefaultModalTransaction()
+    else setConfirmationModalProperties(newProperties)
+  }
```
robbiecarlton: I'm hoping we can avoid this function and just use `setConfirmationModalProperties`.

src/holofuel/pages/Inbox/Inbox.js
robbiecarlton: also, afaict we're not actually using `hasConfirmed` at all

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -323,16 +327,16 @@ export function TransactionRow ({ transaction, setActionsVisibleId, actionsVisib
...
-  let highlightGreen, highlightRed, highlightNeutral
-  if (disabledTransaction) {
-    highlightGreen = increaseAction()
-    highlightRed = decreseAction()
-    highlightNeutral = statusQuoAction()
-  }
+  const findTransactionById = (transactionId, arrayOfTransactions) => arrayOfTransactions.find(tx => tx.id === transactionId)
+
+  // NB: once modalActionedTransaction no longer has an array or transactions, we can update to:  disabledTransaction = transaction.id === modalActionedTransaction.transaction.id
+  const disabledTransaction = findTransactionById(transaction.id, modalActionedTransaction)
```
robbiecarlton: so this is cheating :)

The idea behind the refactor was that the TransactionRow itself is responsible for storing the state it needs to know how to display. Here you're passing that in through the `confirmationModalProperties`.

`ConfimationModal` should be the only consumer of `confirmationModalProperties`

robbiecarlton (changes requested):

robbiecarlton:
@JettTech let's pair on this

src/holofuel/pages/Inbox/Inbox.js
JettTech: We're passing a list in the `DeclinedTransactionModal` so we can resuse the same function to update the actined-transaction in the row, as the only difference *is* that this one contains a list of tx.

Once we rid of this modal, we can simplify the functionality.

src/holofuel/pages/Inbox/Inbox.js
JettTech: We had a prop in the modals called `handleClose` before, which would call this same state hook and set it to null.  Since we're now using that same state hook to update the actioned-transaction row itself, I decided to only pass in this setState hook once instead of twice (once explicitly for `handleClose `and the other for updating the state with the params).

src/holofuel/pages/Inbox/Inbox.js
JettTech: At first, I was too.  ...but the way that our whole component is written has extracted the zome call functionality into the modals.  Consequently, the knowledge of which transaction has been action-ed lives in the modals, which is a sibling of `TransactionRow`, where each of the transactions live. This left some pretty messy alternatives imo, but I'm happy to pair on this though... ! :)

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -27,36 +43,25 @@ export function PrimaryLayout ({
...
-      goToInbox()
+    if (!isEmpty(declinedOffers)) {
+      refundTransactions(declinedOffers)
     }
-  }, [filterActionableTransactionsByStatusAndType, goToInbox])
-
-  const childrenWithProps = React.Children.map(children, child => {
-    if (!isEmpty(child)) return React.cloneElement(child, { whoami: holofuelUser })
-  })
+  }, [refundTransactions, declinedOffers])
```
JettTech, line 64: So happy about his approach!

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -323,16 +267,52 @@ export function TransactionRow ({ transaction, setActionsVisibleId, actionsVisib
...
-  const disabledTransaction = disableActionedTransaction(transaction.id)
-  let highlightGreen, highlightRed, highlightNeutral
-  if (disabledTransaction) {
-    highlightGreen = increaseAction()
-    highlightRed = decreseAction()
-    highlightNeutral = statusQuoAction()
+  const [highlightGreen, setHighlightGreen] = useState(false)
+  const [highlightRed, setHighlightRed] = useState(false)
+  const [isDisabled, setIsDisabled] = useState(false)
+  const [isVisible, setIsVisible] = useState(true)
```
JettTech, line 273: Can we use this same trick in the transaction history page?  Such that the row disappears whenever we receive back a successful mutation?

JettTech (commented):
This really morphed in to a strong update ticket!  Happy we got these odds and ends done!

I noticed one update think should be apart of this ticket - hiding the transaction row in transaction history whenever a successful cancellation is complete as the page preforms the refetch completed transactions call.

Also, I'd like to discuss how we want to manage the display of the transaction row labels whenever we hide the last transaction in any date label.  I fine making this a separate ticket, but I do think we should address it sooner than later (ideally before the first release).

PR #191, opened 2020-02-23
remove fees from calculation for amount returned to avail balance onc…

JettTech:
…e offered payment is declined


zo-el (approved):

robbiecarlton (approved):
LGTM :shipit:

PR #193, opened 2020-03-02
0732 holofuel cleanup no cancel

JettTech:


src/holofuel/pages/TransactionHistory/TransactionHistory.test.js
```diff
@@ -344,7 +346,8 @@ describe('TransactionHistory', () => {
       expect(cancelPendingRequestMock.newData).toHaveBeenCalled()
     })
 
-    it('should open CancellationModal and trigger HolofuelCancelMutation for Pending Offer', async () => {
+    // NB: Intentionally skipped all Cancellation funcationality until DNA/Core cancellation bug is resolved.
```
robbiecarlton: can you add something searchable to all these comments so that we can find them. how about 
`// CANCEL-BUG: Intentionally skipped all Cancellation funcationality until DNA/Core cancellation bug is resolved.`

src/pages/Settings/Settings.js
```diff
@@ -130,12 +131,13 @@ export function Settings ({ history: { push } }) {
         bottomStyle
         dataTestId='network-type'
         value={!isEmpty(status) && status.networkId ? presentHash(status.networkId, 14) : 'Not Available'} />
-      <div styleName='settings-header'>Access Port Numbers</div>
+      {/* NB: Intentionally commented out for now. Defer to Holo Central. */}
```
robbiecarlton: let's just remove this. Keeping the cancel stuff makes sense because thats waiting on 1 specific bug fix. Who knows when/if we'll ever need this ports stuff.

robbiecarlton (changes requested):

src/holofuel/pages/TransactionHistory/TransactionHistory.test.js
```diff
@@ -166,7 +166,8 @@ describe('TransactionHistory', () => {
       expect(queryByText(pendingOutgoingAmount)).not.toBeInTheDocument()
 
       expect(getByText(completedOutgoing.notes)).toBeInTheDocument()
-      expect(getByText(presentHolofuelAmount(completedOutgoing.presentBalance))).toBeInTheDocument()
+      //  NB: Intentionally commented out until DNA balance bug is resolved
```
robbiecarlton: missed a BALANCE-BUG

robbiecarlton (changes requested):
Just one last comment to tag and then it's good to merge. No need for me to review again

PR #196, opened 2020-03-03
7433 - As an HP Admin user I can rename my HoloPort and see the updated name upon saving, without having to reload my browser

robbiecarlton:


src/pages/Settings/Settings.js
```diff
@@ -46,11 +46,14 @@ export function Settings ({ history: { push } }) {
...
       variables: {
-        ...settings,
+        ...pick(['hostPubKey', 'hostName', 'sshAccess'], settings),
         deviceName: editedDeviceName
-      }
+      },
+      refetchQueries: [{
+        query: HposSettingsQuery
+      }]
+
```
JettTech, line 55: Good catch. :+1:

JettTech (approved):
Just fix the small linting complaint, but otherwise LGTM!  Nice test additions. :shipit:

PR #200, opened 2020-03-05
5759 - merge develop into release v0.10.0

robbiecarlton:


src/holofuel/RefundDeclinedOffers.js
```diff
@@ -32,17 +32,15 @@ function RefundDeclinedOffers ({
...
+  const toBeRefundedIds = difference(declinedOffers.map(declinedOffer => declinedOffer.id), refundedTransactionIds)
 
   useEffect(() => {
     if (!isEmpty(toBeRefundedIds)) {
-      refundTransactions(toBeRefundedIds)
-        .then(() => {
-          setRefundedTransactionIds(refundedTransactionIds => refundedTransactionIds.concat(toBeRefundedIds))
-        })
+      refundTransactions(declinedOffers)
+      setRefundedTransactionIds(refundedTransactionIds.concat(toBeRefundedIds))
```
robbiecarlton, line 42: this change means we are updating the local state before we know whether the call succeeded. So if there is a zome error, the transactions will still be marked as refunded in the UI  so the UI will stop trying to refund them.

JettTech: True, but the promise.then was causing major issues, so I took it out for the time being.  Happy to look into why today.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -68,44 +67,52 @@ function useDecline () {
...
+    }
+
+    if (isLoadingFirstTime && !loading) {
+      setHasLoadedOnce(true)
+    }
+  }, [loading, isLoadingFirstTime, setIsLoadingFirstTime, hasLoadedOnce, setHasLoadedOnce])
+
+  return !hasLoadedOnce && loading
+}
+
```
robbiecarlton, line 95: @JettTech this is the change so that the loaders only show the first time.

PR #204, opened 2020-03-08
updated flash message to reflect app context

JettTech:


JettTech:
merging into master this time

PR #208, opened 2020-03-13
Update holo-nixpkgs to use HC v0.0.45-alpha5

zo-el:
### Updates:
- Updated HC to v0.0.45-alpha1
- Updated agent keys to use null passphrase key.
> I don't know why I didn't do this earlier


JettTech (approved):
LGTM. Let's get this into the release to support local hf use as well.

PR #210, opened 2020-03-15
Release v0.10.0 alpha3

JettTech:
- adds the wrapper which automatically calls acceptOffer once an payment is offered in response to a request
- updates holofuel version


philipbeadle (approved):
Looks good, only advice is to write some tests for the fail cases.

zo-el (approved):

JettTech:
@philipbeadle - great advice.  I've made a PT ticket to represent that work. :+1:

PR #214, opened 2020-03-19
Rename conductors and use local networking by default

robbiecarlton:
Hey guys, made some changes to some of our manual testing setup. Please review carefully as I don't know what I'm doing here and might be breaking important stuff (especially integration with `hpos`, which I've not tested). I tested `Holofuel` locally with this setup and it worked fine.


default.nix
```diff
@@ -73,8 +73,8 @@ let
     else [x];
 
   dnas = with dnaPackages; [
-    happ-store
-    holo-hosting-app
+#    happ-store
+#    holo-hosting-app
```
robbiecarlton, line 77: I can imagine this change being especially controversial. How do you feel about this? We're not using them currently right? It just saves me having to have my own copy of `default.nix` around all the time, and the risk of accidentally committing it etc.

default.nix
```diff
@@ -200,7 +200,7 @@ in
       }
     ];
     network = {
-      sim2h_url = "ws://sim2h-test.holo.host:9003";
+      sim2h_url = "ws://localhost:9000";
```
robbiecarlton, line 203: if these conductors are only used for our manual testing, seems like we should default to them being on a local network.

shell.nix
```diff
@@ -15,8 +15,8 @@ mkShell {
   shellHook = ''
     rm -f conductor-config.toml
     rm -f conductor-config-extra.toml
-    ln -s ${project.hp-admin-conductor-config} conductor-config.toml
-    ln -s ${project.hp-admin-extra-conductor-config} conductor-config-extra.toml
+    ln -s ${project.hp-admin-conductor-config-1} conductor-config-1.toml
+    ln -s ${project.hp-admin-conductor-config-2} conductor-config-2.toml
```
robbiecarlton, line 19: For some reason these two lines complain when I run `nix-shell` the second and subsequent times

```
ln: failed to create symbolic link 'conductor-config-1.toml': File exists
ln: failed to create symbolic link 'conductor-config-2.toml': File exists
```

Even though all I changed was the file names and it doesn't complain in the current setup. Either of you have any ideas where I went wrong with this?

.gitignore
```diff
@@ -35,5 +35,5 @@ conductor.log
 conductor-1.log
 conductor-2.log
 conductor.pid
-conductor-config.toml
-conductor-config-extra.toml
+conductor-config-1.toml
+conductor-config-2.toml
```
robbiecarlton, line 39: These names seems clearer to me.

shell.nix
```diff
@@ -15,8 +15,8 @@ mkShell {
   shellHook = ''
     rm -f conductor-config.toml
     rm -f conductor-config-extra.toml
```
zo-el:
```suggestion
    rm -f conductor-config-1.toml
    rm -f conductor-config-2.toml
```

shell.nix
zo-el: The suggestion should help fix this issue

default.nix
zo-el: If we are clear that the `holo-hosting-app` zome calls that will be made by the `hp-admin` api will not go through and we can agree on it, I am **ok** with this change.
(At this stage of where we are at).

I do not know if you have already done this but I would also suggest you make a `Issue` on the `holo-nixpkgs` repo for the build issue that you are facing on a mac.

zo-el (approved):
I am ok with these changes. 
Only concerned about removing the `holo-hosting-app` from the build but if we agree with it, I am good.
  
Just one fix suggested to the issue you were having and its good to go.

shell.nix
JettTech: @robbiecarlton - Try running `realpath /path/file` for both to see where each file lives.  

After locating them, delete those files, exit nix-shell, delete the existing conductor-configs and then re-enter into nix to trigger the nix setup with the desired updates.

.gitignore
JettTech: Been wanting to do this for a  while too.  :+1:

JettTech (changes requested):
Firstly, let's get these tests passing.

Secondly, I agree with most of these changes, but would like to discuss the two following and come to a team agreement on:
1.  whether we'd like to reference the local sim2h as a default instead of our private happy sim2h.
2.  whether we'd like to comment out the `hha` and `happ-store` as a consequence of holo-nixpkgs builds not working with macs.  => This seems like we would be creating another band-aid that covers up the actual issue and doesn't incentivize us to work with holo central to resolve it and/or create better patterns for coordination forward-going.

default.nix
JettTech: @robbiecarlton - I second @joel on the fact that we should leave an issue on holo-nixpkgs with details about your machine and error log, so they can create a build that accommodates 86_64 Darwin OS (with Catalina).

shell.nix
robbiecarlton: @zo-el I don't see the suggestion

PR #215, opened 2020-03-19
6946 - Edit Nickname

JettTech:


README.md
```diff
@@ -39,7 +39,7 @@ If referencing **Live** Data :
 **NOTE: Please reference version v0.12.2 of the holofuel dna to test with the UI.**
 - Open 2 terminals.
 ##### Start the Conductors:
-- In one of the termainals, open nix-shell by typing `nix-shell` (after this happens your terminal line should become neon green)
+- In one of the terminals, open nix-shell by typing `nix-shell` (after this happens your terminal line should become neon green)
```
robbiecarlton, line 42: 👍

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -197,26 +197,38 @@ function presentTransaction (transaction) {
 const HoloFuelDnaInterface = {
   user: {
     get: async () => {
-      const result = await createZomeCall('transactions/whoami')()
-      if (result.Err) throw new Error('There was an error locating the current holofuel agent nickname. ERROR: ', result.Err)
-
+      const myProfile = await createZomeCall('profile/get_my_profile')()
+      if (myProfile === 'Err') throw new Error('There was an error locating the current holofuel agent profile. ERROR: ', myProfile)
```
robbiecarlton: shouldn't this be
`if (myProfile.Err)` 
or similar? Does the endpoint just return the string `Err` on error?

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -197,26 +197,38 @@ function presentTransaction (transaction) {
...
       return {
-        id: result[0].Ok.agent_id.pub_sign_key,
-        nickname: result[0].Ok.agent_id.nick
+        id: counterpartyProfileArray[0].agent_address,
+        avatarUrl: counterpartyProfileArray[0].avatar_url,
+        nickname: counterpartyProfileArray[0].nickname
+      }
+    },
+    update: async (id, nickname, avatarUrl) => {
+      const params = avatarUrl ? { nickname, avatar_url: avatarUrl } : { nickname }
```
robbiecarlton: better to do this using lodash omitBy
`const params = omitBy(param => param === undefined, { nickname, avatarUrl })`

(eg, this approach handles the case where avatarUrl is supplied and nickname isn't)

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -197,26 +197,38 @@ function presentTransaction (transaction) {
...
+      const myProfile = await createZomeCall('profile/update_my_profile')(params)
+      if (myProfile === 'Err') throw new Error('There was an error udpating the current holofuel agent profile. ERROR: ', myProfile)
```
robbiecarlton: same question about the `'Err'` string

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -286,7 +297,8 @@ const HoloFuelDnaInterface = {
       const nonActionableTransactions = transactions.map(presentTransaction)
       const noDuplicateIds = _.uniqBy(nonActionableTransactions, 'id')
 
-      const whoami = await HoloFuelDnaInterface.user.get()
+      const myProfile = await HoloFuelDnaInterface.user.get()
+      const whoami = myProfile.id
```
robbiecarlton: seems like a type change here. In the old code, `whoami` was an object with two fields, `id` and `nickname`. Now it's a string (that's the `id` of the user).

But the consuming code us not changed (see line 305 below (in the new code)). This is passing an id as `canceledBy`, but canceledBy should be an object.

src/graphql-server/resolvers.js
```diff
@@ -89,6 +89,8 @@ export const resolvers = {
 
     updateHostPricing: (_, { units, pricePerUnit }) => HhaDnaInterface.hostPricing.update(units, pricePerUnit),
 
+    holofuelUpdateUser: (_, { id, nickname, avatarUrl }) => HoloFuelDnaInterface.user.update(id, nickname, avatarUrl),
```
robbiecarlton: we can only update ourselves so no need to pass the `id` here

src/graphql-server/schema.graphql
```diff
@@ -152,6 +153,7 @@ type Mutation {
   enableHapp(appId: String): Happ
   disableHapp(appId: String): Happ
   updateHostPricing(units: String, pricePerUnit: String): HostPricing
+  holofuelUpdateUser(id: String, nickname: String!, avatarUrl: String): HolofuelUser
```
robbiecarlton: we can only update ourselves so no need for `id`

src/graphql/HolofuelUpdateUserMutation.gql
```diff
@@ -0,0 +1,7 @@
+mutation HolofuelUpdateUser ($id: String, $nickname: String!, $avatarUrl: String) {
```
robbiecarlton: no need for id

src/holofuel/UpdateHolofuelUser.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState, useEffect } from 'react'
```
robbiecarlton: can we call this component `PromptForNickname` or similar. `UpdateHolofuelUser` makes me think it's the page where we do the updating

src/holofuel/UpdateHolofuelUser.js
```diff
@@ -0,0 +1,34 @@
...
+}) {
+  const { data: { holofuelUser: { id, nickname } = {} } = {} } = useQuery(HolofuelUserQuery)
+  const [hasReceivedNotice, setHasReceivedNotice] = useState(false)
+
+  const { newMessage } = useFlashMessageContext()
+  const history = useHistory()
+  const pathname = history.location.pathname
+
+  useEffect(() => {
+    if (hasReceivedNotice) return
```
robbiecarlton: :'( 
I know we need it, but still
:'(

src/holofuel/UpdateHolofuelUser.js
```diff
@@ -0,0 +1,34 @@
...
+    if (pathname === `/holofuel/${PROFILE_PATH}`) {
+      newMessage('', 0)
+    } else if (!isEmpty(id) && isEmpty(nickname)) {
+      newMessage(`It looks like you don't yet have a Holofuel Account Nickname. Visit your profile page to personalize your nickname so your peers can easily recognize you.`, 0, 'Click here to visit your profile.', PROFILE_PATH)
```
robbiecarlton: I think newMessage should just be able to take arbitrary JSX to render as the first param. Then you don't have the awkward interface of passing in an optional link title and path

src/holofuel/components/FlashMessage/FlashMessage.js
```diff
@@ -20,7 +21,14 @@ export default function FlashMessage () {
...
   return <div styleName='flash-message'>
-    <div styleName='message'>{message}</div>
+    <div styleName='flash-body'>
+      {link ? <>
+        <div styleName='message'>{message}</div>
+        <br />
+        <Link to={link}>{linkText}</Link>
+      </>
+        : <div styleName='message'>{message}</div> }
+    </div>
```
robbiecarlton: I'd like to handle this differently.

Instead of having a specific `linkText` and `link` that you can pass in, it should just test if `message` is a string. If so, render it in a div, if not, assume it's jsx and render it directly.

Actually, if you always wrap it in the `'message'` div, you don't even need to check.

src/holofuel/components/Header/Header.js
```diff
@@ -25,9 +26,9 @@ export function Header ({ title, agent, agentLoading, avatarUrl, history: { push
         </div>}
       </div>
       <div>
-        <CopyAgentId agent={agent} isMe>
-          <HashAvatar avatarUrl={avatarUrl} seed={agent.id} size={32} dataTestId='hash-icon' />
-        </CopyAgentId>
+        <Link to={PROFILE_PATH}>
+          <HashAvatar seed={agent.id} size={32} dataTestId='hash-icon' />
+        </Link>
```
robbiecarlton: I think we want to keep the copy functionality. Not sure. We went back and forth on this one and not sure where we landed. But I missed it when testing

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -18,6 +17,7 @@ import 'holofuel/global-styles/index.css'
 
 function PrimaryLayout ({
   children,
+  color,
```
robbiecarlton: doesn't look like you're using this prop?

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.module.css
```diff
@@ -38,3 +38,7 @@
 .side-menu {
   padding-right: 10px;
 }
+
+.iceburg {
```
robbiecarlton: sp: iceberg 
;)

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -77,10 +77,12 @@ export default function CreateOfferRequest ({ history: { push } }) {
 
   useEffect(() => {
     setCounterpartyNick(presentAgentId(counterpartyId))
-    if (counterpartyId === whoami.id) {
+    // console.log('counterpartyId === myProfile.id : ', counterpartyId === myProfile.id)
```
robbiecarlton: remove comment

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -77,10 +77,12 @@ export default function CreateOfferRequest ({ history: { push } }) {
...
+
+    if (counterpartyId === myProfile.id) {
       newMessage('You cannot send yourself TestFuel.', 5000)
     }
-  }, [whoami.id, counterpartyId, newMessage])
+  }, [myProfile, counterpartyId, newMessage])
```
robbiecarlton: dependency can still be `myProfile.id`

src/holofuel/pages/Profile/Profile.js
```diff
@@ -0,0 +1,62 @@
...
+        query: HolofuelUserQuery
+      }]
+    })
+    reset({ nickname: '' })
+  }
+
+  return <PrimaryLayout headerProps={{ title: 'Profile' }}>
+    <div styleName='backdrop' />
+    {loading && <Loading />}
+    <Card title='Update Profile' subtitle='Manage your account nickname and avatar.'>
```
robbiecarlton: should just be `'Manage your account nickname'`

robbiecarlton (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Agreed. Good catch.

src/graphql-server/resolvers.js
JettTech: This is just so we can return the full profile back, once completing a successful zome call, as the dna returns an address.

src/graphql-server/schema.graphql
JettTech: Same reason as above.

src/graphql/HolofuelUpdateUserMutation.gql
JettTech: Same reason as above.

src/graphql-server/resolvers.js
JettTech: Ok. We decided that we will extract this id from the UI data layers and just make a separate call to the DNA to get the agent hash, in order to clean the UI and prepare for the DNA update that returns back the full profile instead of just the hash address.

src/holofuel/UpdateHolofuelUser.js
JettTech: haha. Agreed.

src/holofuel/UpdateHolofuelUser.js
JettTech: I was actually just thinking through that again prior to our review call and updated it to `updateHolofuelUserPrompt`. Thoughts?

src/holofuel/components/Header/Header.js
JettTech: This is definitely a UX decision.  I'm fine with either implementation for now.

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: This was residual for a non-final design decision... good catch.

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.module.css
JettTech: haha. Very true. I actually just copied the title of color var already created in our global styles without really reading over it. I'll go ahead and do a global update. :+1:

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -197,26 +197,39 @@ function presentTransaction (transaction) {
 const HoloFuelDnaInterface = {
   user: {
     get: async () => {
-      const result = await createZomeCall('transactions/whoami')()
-      if (result.Err) throw new Error('There was an error locating the current holofuel agent nickname. ERROR: ', result.Err)
-
+      const myProfile = await createZomeCall('profile/get_my_profile')()
+      if (myProfile.Err) throw new Error('There was an error locating the current holofuel agent profile. ERROR: ', myProfile.Err)
```
robbiecarlton: Error doesn't concatenate strings, so you need to use a template literal
`There was an error locating the current holofuel agent profile. ERROR: ${myProfile.Err}`

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -197,26 +197,39 @@ function presentTransaction (transaction) {
...
+        id: counterpartyProfileArray[0].agent_address,
+        avatarUrl: counterpartyProfileArray[0].avatar_url,
+        nickname: counterpartyProfileArray[0].nickname
+      }
+    },
+    update: async (nickname, avatarUrl) => {
+      const { id } = await HoloFuelDnaInterface.user.get()
+      const params = omitBy(param => param === undefined, { nickname, avatarUrl })
+      const myProfile = await createZomeCall('profile/update_my_profile')(params)
+      if (myProfile.Err) throw new Error('There was an error udpating the current holofuel agent profile. ERROR: ', myProfile.Err)
```
robbiecarlton: use a template string

src/holofuel/UpdateHolofuelUserPrompt.js
```diff
@@ -0,0 +1,48 @@
...
+  const userMessage = `It looks like you don't yet have a Holofuel Account Nickname. Visit your profile page to personalize your nickname so your peers can easily recognize you.`
+  const linkMessage = `Click here to visit your profile.`
+  return <>
+    <div className='message'>{userMessage}</div>
+    <br />
+    <Link to={PROFILE_PATH} onClick={handleResetFlashMessage}>{linkMessage}</Link>
+  </>
+}
+
+function UpdateHolofuelUserPrompt ({
```
robbiecarlton: Still like `PromptForNickname`

I like using a verb to name these non rendering components, because they are DOING something, not BEING something. Eg `AcceptRequestOffers`

src/holofuel/UpdateHolofuelUserPrompt.js
```diff
@@ -0,0 +1,48 @@
...
+import { useQuery } from '@apollo/react-hooks'
+import { isEmpty } from 'lodash/fp'
+import HolofuelUserQuery from 'graphql/HolofuelUserQuery.gql'
+import useFlashMessageContext from 'holofuel/contexts/useFlashMessageContext'
+import { PROFILE_PATH } from 'holofuel/utils/urls'
+import 'holofuel/global-styles/colors.css'
+import 'holofuel/global-styles/index.css'
+
+function UserPromptMessage ({ newMessage }) {
+  const handleResetFlashMessage = useCallback(
```
robbiecarlton: nitpicking (and not blocking), but we can just call this `resetFlashMessage`

`handle` implies this is a callback that is called WHEN the message is reset, rather than the function which is actually responsible for resetting the message,

robbiecarlton (changes requested):

src/holofuel/UpdateHolofuelUserPrompt.js
JettTech: Good call. I support this thinking; updating it now.

zo-el (approved):

PR #216, opened 2020-03-24
6946 profile updates & 1685 fix create profile display latency bug

JettTech:


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -197,39 +197,31 @@ function presentTransaction (transaction) {
 const HoloFuelDnaInterface = {
   user: {
     get: async () => {
-      const myProfile = await createZomeCall('profile/get_my_profile')()
-      if (myProfile.Err) throw new Error(`There was an error locating the current holofuel agent profile. ERROR: ${myProfile.Err}.`)
+      const { agent_address: id, avatar_url: avatarUrl, nickname, Err } = await createZomeCall('profile/get_my_profile')()
+      if (Err) throw new Error(`There was an locating your holofuel agent ID. ERROR: ${Err}.`)
```
robbiecarlton: you accidentally a word

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -197,39 +197,31 @@ function presentTransaction (transaction) {
...
       return {
-        id: myProfile.agent_address,
-        avatarUrl: myProfile.avatar_url,
-        nickname: myProfile.nickname
+        id,
+        avatarUrl: avatarUrl,
+        nickname: nickname
```
robbiecarlton: not important but you can do the compact syntax here
```
avatarUrl,
nickname
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -197,39 +197,31 @@ function presentTransaction (transaction) {
...
       const counterpartyProfileArray = await createZomeCall('profile/get_profile')({ agent_address: agentId })
-      if (counterpartyProfileArray === 'Err' || !counterpartyProfileArray[0]) {
-        return {
-          id: agentId,
-          avatarUrl: '',
-          nickname: '',
-          notFound: true
-        }
-      }
+      if (counterpartyProfileArray.Err) throw new Error(`There was an locating the holofuel agent with ID: ${agentId}. ERROR: ${counterpartyProfileArray.Err}. `)
```
robbiecarlton: missed the word 'error' in this message.

Also, should this variable be called "array". Seems like it's just a `counterpartyProfile`

robbiecarlton: Also, it seems like this shouldn't throw the error. We're not catching it anywhere? So I think we should "catch" it here, and just return 
```
{
  id: null,
  avatarUrl: null,
  nickname: null
}
```
when `Err` is set

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -418,14 +419,18 @@ export function ConfirmationModal ({ confirmationModalProperties, setConfirmatio
   const { newMessage } = useFlashMessageContext()
   const { transaction, action, shouldDisplay, onConfirm, setIsLoading } = confirmationModalProperties
 
+  console.log('confirmationModalProperties : ', confirmationModalProperties)
```
robbiecarlton: remove log

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -418,14 +419,18 @@ export function ConfirmationModal ({ confirmationModalProperties, setConfirmatio
...
+
   const { id, amount, type, notes, counterparty = {} } = transaction
   const { loading: loadingCounterparty, holofuelCounterparty } = useCounterparty(counterparty.id)
-  const { notFound } = holofuelCounterparty
+  const { id: activeCounterpartyId } = holofuelCounterparty
+
+  console.log('holofuelCounterparty : ', holofuelCounterparty)
```
robbiecarlton: remove log

src/holofuel/pages/Profile/Profile.js
```diff
@@ -43,7 +45,7 @@ export default function Profile () {
         <CopyAgentId agent={{ id }} isMe>
           <HashAvatar seed={id} styleName='avatar-image' data-testid='host-avatar' />
         </CopyAgentId>
-        <h3 styleName='nickname-display' data-testid='profile-nickname'>{nickname || 'Your Nickname'}</h3>
+        <h3 styleName='nickname-display' data-testid='profile-nickname'>{nickname || optimisitcNickname || 'Your Nickname'}</h3>
```
robbiecarlton: `optimisticNickname` should override the `nickname` that comes from the query, so it should appear first in the `||` chain.

robbiecarlton (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: I love the way you really emphasized by example! :joy:

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: I also prefer to maintain the error catching/handling in the dna logic, and I also think we should return an error to the console to demarcate undesired/unexpected behavior.  Therefore, I'll return the null values as well as add a console.error.

src/holofuel/pages/Inbox/Inbox.js
JettTech: :man_facepalming: Thanks for calling out the cleanup.

src/holofuel/pages/Profile/Profile.js
JettTech: This is true. Updating now.

robbiecarlton (approved):

robbiecarlton (approved):

src/holofuel/pages/Profile/Profile.js
```diff
@@ -20,10 +20,16 @@ function Card ({ title, subtitle, children }) {
...
-  const { loading, data: { holofuelUser: { id, nickname } = {} } = {} } = useQuery(HolofuelUserQuery, { fetchPolicy: 'cache-and-network' })
+  const { loading, data: { holofuelUser: { id, nickname } = {} } = {}, refetch: refetchHolofuelUser } = useQuery(HolofuelUserQuery, { fetchPolicy: 'cache-and-network' })
   const [updateUser] = useMutation(HolofuelUpdateUserMutation)
   const [optimisitcNickname, setOptimisitcNickname] = useState('')
 
+  useEffect(() => {
+    if (optimisitcNickname && !nickname) {
+      refetchHolofuelUser()
+    }
+  }, [optimisitcNickname, nickname, refetchHolofuelUser])
```
robbiecarlton: this might cause an infinite loop if for some reason the nickname didn't make it into the backend. Let's discuss an approach.

robbiecarlton (changes requested):

src/holofuel/pages/Profile/Profile.js
```diff
@@ -32,6 +42,9 @@ export default function Profile () {
         query: HolofuelUserQuery
       }]
     })
+      .catch(() => setOptimisticNickname())
```
robbiecarlton, line 45: @JettTech one other addition for your attention, here I'm catching any potential error from the update call. If it throws an error, we reset the optimisticNickname to what it was on page load.

PR #218, opened 2020-03-24
3694 - As a user I can see the version of HP Admin on the login screen so that debugging and troubleshooting is easier.

robbiecarlton:


JettTech (changes requested):
All updates looks good; let's just get those tests to pass.  :+1:

PR #224, opened 2020-03-25
7538 - HP-ADMIN optimization: 'CurrentUser' context

JettTech:
- create whoami context for hp-admin
- replaces redundant hp-admin call to hpos settings tor host's info with hp-admin whoami context
- update naming to match holofuel whoami context naming pattern
- correct spelling error in holofuel pre-seed script


scripts/prepareHoloFuelData/transact-holofuel.js
```diff
@@ -20,7 +20,7 @@ const transactHoloFuel = (agentTransactionLedger, type, { index, transactionTrac
     currentAgent = 1
   }
 
-  const txType = type === OFFER ? agentTransactionLedger[`offers`][`initated`] : agentTransactionLedger[type]
+  const txType = type === OFFER ? agentTransactionLedger.offers.initated : agentTransactionLedger[type]
```
Connoropolous: 👋  hi @JettTech 
forgive me for saying something here, I know you're busy. I have been keeping up with various Holo and Holochain PRs, and watching along, and I spotted one thing here that seemed wrong not to point out, although it's possible it's already known. 

In this line it says `.offers.initated` and based on how it was before, that is correct in terms of the data, but `initated` should `initiated` if the word were spelled right. Anyways this seemed like the kind of thing that could easily cause a bug now or somewhere down the line, so it might be worth fixing in the upstream dependency. 
🤷‍♂ 
carry on, and feel free to ignore if this doesn't make sense.

JettTech: Good eye @Connoropolous! :eye:   ...and no, it has not been surfaced yet, so thanks for pointing that out!  Sometimes, I wish I could have a grammerly plugin for my code editor sometimes to manage typos! :joy_cat:  Nonetheless, great to know your scrupulous eyes are around... they always provide great support! :raised_hands:

src/components/icons/LaptopIcon.js
```diff
@@ -7,7 +7,8 @@ export default function LaptopIcon ({ className, color = '#000000' }) {
   return <Icon
     color={color}
     title={title}
-    className={className}>
+    className={className}
+  >
```
robbiecarlton, line 11: I don't like this change. Is the linter making you do it? I don't think the version of standardjs we use makes us do this, but maybe a newer version does?

src/graphql-server/schema.graphql
```diff
@@ -106,6 +106,7 @@ type HposStatus {
 
 type HposAuth {
   isAuthed: Boolean
+  hostSettings: HposSettings
```
robbiecarlton: can we just call this field `hposSettings`
in less there's a specific reason not to, it's good to name your variables the same name as their type.

src/holofuel/components/HashIcon/HashIcon.js
```diff
@@ -13,7 +13,8 @@ export default function HashIcon ({
       seed={hash}
       size={size}
       className={className}
-      circle />
+      circle
+    />
```
robbiecarlton: really don't want us to do this if we can avoid it. I think this is a standardjs version thing. Your linter might be a newer version of standard than the one in the repo

robbiecarlton (changes requested):
Let's get on the same page about the linting thing. If we can avoid newlines for those closing angle brackets, I'd be really happy.

Otherwise, looks good.

src/components/icons/LaptopIcon.js
JettTech: Updated (... or downgraded? :joy: ) to 13.1.0. :+1:

src/graphql-server/schema.graphql
JettTech: The type name was made to match/reflect the endpoint name, but I agree that the gql type should match the var name.  Update done. :+1:

src/holofuel/components/HashIcon/HashIcon.js
JettTech: standard 13.1.0 does the trick. :tada: :+1:

src/graphql-server/schema.graphql
```diff
@@ -106,6 +111,7 @@ type HposStatus {
 
 type HposAuth {
   isAuthed: Boolean
+  hostSettings: HostSettings
```
robbiecarlton: I meant: call this variable `hposSettings`, so the whole line would be `hposSettings: HposSettings`.

I don't think we need to add a new type here, think we can just reuse `HposSettings`.

Can you rename it, and remove the newly added `HostSettings` type?

Other than that, this PR is good to go. Feel free to merge once this change is made.

robbiecarlton (changes requested):

src/graphql-server/schema.graphql
JettTech: Got it.  Agree.  Updated.

PR #227, opened 2020-03-26
(do not merge) 3579 - HoloFuel Optimization : create`CurrentUser` context

robbiecarlton:


JettTech (approved):
Looks good to me, almost identical to the code that was done in the cleanup whoami branch (as we expected), but I do like your naming better. :smiley:

Anyway, for purposes of keeping together the updates to whoami for both holofuel and hp-admin, I will update the merge base for this branch to the cleanup whoami branch, which also includes the hp-admin whoami context wrapper - that way we can have a clean commit for all curentUser/whoami updates.

JettTech:
:+1:

PR #228, opened 2020-03-26
Hot-fix: handle error instead of throwing it

zo-el:
- Updated the holo-nixpkgs 
- handle errors in the list or transactions.


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -249,7 +248,7 @@ const HoloFuelDnaInterface = {
       const params = since ? { since } : {}
 
       const { transactions } = await createZomeCall('transactions/list_transactions')(params)
-      const nonActionableTransactions = transactions.map(presentTransaction)
+      const nonActionableTransactions = transactions.map(presentTransaction).filter(tx => !(tx instanceof Error))
```
robbiecarlton: think you want to filter the list before mapping `presentTransaction` over it. `presentTransaction(anError is not defined)`

robbiecarlton, line 251: think you want to filter the list before mapping `presentTransaction` over it. `presentTransaction(anError)` is not defined

robbiecarlton (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
robbiecarlton: obv, same feedback everywhere you're using this pattern

zo-el: @robbiecarlton, I dont think I understand. 
According to me the map build the array with the errors and then we filter it.

robbiecarlton: My mistake. I misread it, you're right.

robbiecarlton (approved):

robbiecarlton:
@zo-el LGTM

JettTech (approved):

PR #229, opened 2020-03-26
3579 - HoloFuel Optimization : create`CurrentUser` context

robbiecarlton:


src/root.js
```diff
@@ -12,7 +12,10 @@ import { AuthProvider } from 'contexts/useAuthContext'
 import { FlashMessageProvider } from 'contexts/useFlashMessageContext'
 import HFScreenWidthContext from 'holofuel/contexts/screenWidth'
 import { FlashMessageProvider as HFFlashMessageProvider } from 'holofuel/contexts/useFlashMessageContext'
-import AcceptRequestedOffers from './holofuel/AcceptRequestedOffers'
+import { CurrentUserProvider as HFCurrentUserProvider } from 'holofuel/contexts/useCurrentUserContext'
+import AcceptRequestedOffers from 'holofuel/components/wrappers/AcceptRequestedOffers'
+import LoadCurrentUser from 'holofuel/components/wrappers/LoadCurrentUser'
+
 import PromptForNickname from './holofuel/PromptForNickname'
```
JettTech: I would like to include this in the wrapper folder as well. (This is non-blocking though.)

JettTech (approved):

src/root.js
robbiecarlton: Ah yes, good catch

PR #233, opened 2020-03-30
minor grammar fix for request funds agent placeholder

Connoropolous:
found this while looking at the other stuff :p 

looking forward to playing with this connected to Holoports :)


robbiecarlton:
thanks @Connoropolous - we're going to change 'for' in that code to 'to', for even clearer grammar.

PR #234, opened 2020-04-01
Decline message cases were mixed up

Connoropolous:
There are two things I've done in this PR...

1. fix the thing that I first noticed, which is that the messages for "Decline Confirmation" were mixed up between the "offer" and "request" transaction types. That was a simple switch

2. Made change proposals to the language use around "Accept request for payment". I don't think this has been thought about very hard, despite the code comment. I don't think it's been appreciated just how confusing/misleading this language is from an end-user perspective. It took someone as familiar as me with all of this a good one minute of reading and re-reading the messages to really make sense of them, and know what they were getting at. Not a good start for something as sensitive as sending out a payment. I made a change proposal to the language of "accept request for payment". I propose words such as "Agree", or "Pay", and also to use that language within the confirmation modal, so that it becomes "Agree to pay 124,500 TF to Sam?" No/Yes. 

Of course you can accept the first without the second change request, but I propose this, without some change, is guaranteed to cause user issues down the road.

Accept/Agree related before/after:

Before

Screen Shot 2020-03-31 at 10 05 33 PM

After

Screen Shot 2020-03-31 at 9 55 15 PM

### DIFF
--- a/src/holofuel/pages/Inbox/Inbox.js
+++ b/src/holofuel/pages/Inbox/Inbox.js
@@ -418,8 +418,7 @@ function PayButton ({ showPayModal }) {
     onClick={showPayModal}
     styleName='accept-button'
   >
-    {/* NB: Not a typo. This is to 'Accept Request for Payment' */}
-    <p>Accept</p>
+    <p>Pay</p>
   </Button>
 }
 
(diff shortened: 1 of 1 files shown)


Connoropolous:
@JettTech @robbiecarlton  @SirRobLyon

marycamacho:
We agree. We have been thinking about this significantly in the past few months during our test cycles and the next version that will be deployed in test already has changes and the release of Hosted hApps will bring even more changes to simplify this design.

Connoropolous:
Good to hear :)
I'm not sure what your comment means for this PR. I will make a separate pull request for point 1. of the two points, which is a bug fix. Point 2 is obviously a larger matter, I take it you'll have a process for integrating those changes at your own pace

PR #237, opened 2020-04-01
3423 whois context

JettTech:


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -12,8 +12,20 @@ const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
 
 const MOCK_DEADLINE = '4019-01-02T03:04:05.678901234+00:00'
 
+/* Creates an array of all transactions that inlcude new counterparties for a provided transaction list and counterparty list */
+export const findnewCounterpartiesFromList = (transactionList = [], counterpartyList = []) => {
```
robbiecarlton: when naming array variables, it's simpler to just name them the a plural. That's all the clue we need to get that it's an array.
So I would rename these variables `transactions` and `counterparties`.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -12,8 +12,20 @@ const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
...
+  // eslint-disable-next-line array-callback-return
+  const newCouterpartyList = transactionList.filter(({ counterparty }) => {
```
robbiecarlton: typo in this name, and I would recommend using the array naming convention above. so `newCounterparties`

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -12,8 +12,20 @@ const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
...
+    if (isEmpty(counterpartyList)) return counterparty
```
robbiecarlton: because this function is being passed to filter, it should only ever return `true` and `false` (or other falsey values). It's confusing to return `counterparty` when we're not using the value, just the truthiness of it.

src/data-interfaces/HoloFuelDnaInterface.js
robbiecarlton: although it's not actually an array of counterparties, it's an array of transactions, so maybe `newCounterpartyTransactions`

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -12,8 +12,20 @@ const createZomeCall = instanceCreateZomeCall(INSTANCE_ID)
...
+    const existingCounterparty = counterpartyList.find(counterpartyInList => counterpartyInList.id === counterparty.id)
+    if (!existingCounterparty) return counterparty
+  })
+  const uniqueNewCounterpartyTransactions = _.uniqBy(newCouterpartyList, 'counterparty')
```
robbiecarlton: I'm not sure about this, but I don't think this actually gets you uniqueness. the counterparty field of each transaction object will be a different object, even if counterparty.id is the same.

Anyway, I think I would make uniqueness the responsibility of `getTxCounterparties`. Seems like there's never any value in that function returning duplicate counterparties.

robbiecarlton: Looking at `getTxCounterparties`, it looks like it already de-dupes the list, so no need to do it here.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -249,16 +259,13 @@ const HoloFuelDnaInterface = {
 
       const { transactions } = await createZomeCall('transactions/list_transactions')(params)
       const nonActionableTransactions = transactions.map(presentTransaction).filter(tx => !(tx instanceof Error))
-      const noDuplicateIds = _.uniqBy(nonActionableTransactions, 'id')
-      const presentedCompletedTransactions = await getTxWithCounterparties(noDuplicateIds.filter(tx => tx.status === 'completed'))
-      return presentedCompletedTransactions.sort((a, b) => a.timestamp > b.timestamp ? -1 : 1)
+      const noDuplicateCompletedTransactions = _.uniqBy(nonActionableTransactions, 'id').filter(tx => tx.status === 'completed')
```
robbiecarlton: this is a bit picky, but how about `uniqueCompletedTransactions`?

I think that's easier to read (a positive is easier to read than a negative, and it's shorter)

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -296,15 +301,12 @@ const HoloFuelDnaInterface = {
       const nonActionableTransactions = transactions.map(presentTransaction).filter(tx => !(tx instanceof Error))
       const noDuplicateIds = _.uniqBy(nonActionableTransactions, 'id')
 
-      const myProfile = await HoloFuelDnaInterface.user.get()
+      const whoami = await HoloFuelDnaInterface.user.get()
```
robbiecarlton: I would just call this `currentUser`

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -33,6 +36,21 @@ function PrimaryLayout ({
...
+  const { counterpartyList, setCounterpartyList } = useCounterpartyListContext()
+
+  useEffect(() => {
+    // eslint-disable-next-line
+    if (hasUpdatedCounterpartyList) return
+    else if (!isEmpty(actionableTransactions)) {
+      findnewCounterpartiesFromList(actionableTransactions, counterpartyList)
+        .then(newCounterparties => {
+          setCounterpartyList([...counterpartyList, ...newCounterparties])
+          setHasUpdatedCounterpartyList(true)
```
robbiecarlton: I'm not sure if we want to use context for this. I was hoping we could use the Apollo cache for this. Otherwise we're starting to roll our own cache out of contexts, which we want to avoid. Let's talk about the possibility of using Apollo cache

robbiecarlton (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Good call.  Updating now. :+1:

JettTech:
This approach was replaced in favor of a caching service.  Closing.

PR #239, opened 2020-04-02
standalone bug fix for 'decline' modals

Connoropolous:
same as #234 but standalone, without the transaction accept language changes


Connoropolous:
@JettTech and @robbiecarlton have you two looked at this? 
It's just a simple bug fix. The two cases of 'offer' and 'request' are mixed up!

PR #242, opened 2020-04-09
B-01414 - Holofuel UI Release Cleanup

robbiecarlton:


src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -457,7 +457,7 @@ export function ConfirmationModal ({ confirmationModalProperties, setConfirmatio
       actionParams = { id, amount, counterparty, notes }
       actionHook = payTransaction
       message = <>
-        Accept request for payment of {presentHolofuelAmount(amount)} TF from {counterparty.nickname || presentAgentId(counterparty.id)}?
+        Accept the request and send {counterparty.nickname || presentAgentId(counterparty.id)} {presentHolofuelAmount(amount)} TF?
```
JettTech, line 460: Should 'request' be capitalized here? It is in the wording located in te Version One ticket, so I just want to verify. (I know this is super nitpicky, but the focus of this ticket is rather detailed, so let's just double-check this here.)

JettTech (changes requested):
All looks well, but I have one question to address and one update to include before merging:
  Update: The 'balance' wording on the Inbox page needs to be updated to 'current balance'
  Question: Should the word 'request' be capitalized or not?

src/holofuel/pages/Inbox/Inbox.js
robbiecarlton: yeah I made a judgement call there. Didn't look right to me. I'll run it passed Rob before hitting complete on the ticket.

JettTech: If we're polling, I agree that it's not a proper noun, and therefore shouldn't be capitalized.

JettTech (approved):
LGTM.  I tested locally and the tests pass, it looks like our CI needs to be recalibrated due to memory issues.

PR #248, opened 2020-04-20
3338 - Fix UI Zome Integration Test Harness

robbiecarlton:


robbiecarlton:
@JettTech @zo-el 
Test harness is now in shape, ready for integration tests to be update, which will be separate work outside of this ticket.

To test this:

1. checkout the branch
2. nix-shell
3. npm run test:initialize-test-data

This will 
1. start up a sim2h server
2. start up the test conductor
3. populate the test data
4. run the (currently all skipped) tests

If all that runs without error, then it works :)

you can see the conductor log in `test-conductor.log` and the sim2h log in `sim2h.log`

zo-el:
> @JettTech @zo-el
> Test harness is now in shape, ready for integration tests to be update, which will be separate work outside of this ticket.
> 
> To test this:
> 
> 1. checkout the branch
> 2. nix-shell
> 3. npm run test:initialize-test-data
> 
> This will
> 
> 1. start up a sim2h server
> 2. start up the test conductor
> 3. populate the test data
> 4. run the (currently all skipped) tests
> 
> If all that runs without error, then it works :)
> 
> you can see the conductor log in `test-conductor.log` and the sim2h log in `sim2h.log`

I think you mean `npm run test:integration-with-setup`

zo-el (changes requested):
1. When you rerun `npm run test:initialize-test-data` or `npm run test:integration-with-setup` it will fail. you need to stop the conductor after ever time you preseed the conductor and ever time you run tests.
2. I think we need one test to prove that the testing harness works. We can also run the same test twice to prove that the reset of the conductor works.

I think we need to do the second point(2) which will let us avoid issues like point one(1)

JettTech:
Looked through this and have a few general questions about the approach, let's sync up and review this at the next PL.

Also, we need to look into how the agent keys / conductor sessions are being handled, as the `Public token not found: Can't get CapTokenGrant entry from CAS` error populates every subsequent run of  `test:initialize-test-data`.

JettTech:
This PR is deprecated in favor of the new Tryorama approach.  Closing out.

PR #249, opened 2020-04-25
B-01531 - add callstack caching

JettTech:


test-conductor-config.toml
```diff
@@ -0,0 +1 @@
+/nix/store/59ckizjah6fhi4qv0m0b5gr5ijdm25as-config.toml
```
zo-el, line 1: gitignore this file

test-conductor.log
```diff
@@ -0,0 +1,2362 @@
+Using config path: ./test-conductor-config.toml
```
zo-el, line 1: gitignore this file

zo-el (changes requested):
@JettTech, Looks good to me.
we should probably make a lib for this so that we can apply it to more that one API or if we find some lib that already exists we can use it.

For our current hot fix. this works.

##### TODO: Please add the conductor and log files to the giignore and remove it from this PR

PR #257, opened 2020-04-29
Prevent actioning multiple transactions

robbiecarlton:


CHANGELOG.MD
```diff
@@ -3,6 +3,10 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## unreleased
```
JettTech, line 6: :muscle: :clap:

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -279,7 +285,7 @@ export function TransactionRow ({ transaction, setConfirmationModalProperties, i
   const [highlightGreen, setHighlightGreen] = useState(false)
   const [highlightRed, setHighlightRed] = useState(false)
   const [isDisabled, setIsDisabled] = useState(false)
-  const [isLoading, setIsLoading] = useState(false)
+  const [isLoading, setIsLoadingRaw] = useState(false)
```
JettTech: This is very particular, but can we update the const name to match the change in the function name?  IMO, it helps identify the pairs well.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -369,14 +381,22 @@ export function TransactionRow ({ transaction, setConfirmationModalProperties, i
...
-    <ForwardIcon styleName='forward-icon' color='#2c405a4d' dataTestId='forward-icon' />
+function RevealActionsButton ({ openDrawer, closeDrawer, isDrawerOpen, areActionsPaused }) {
+  const onClick = isDrawerOpen
+    ? closeDrawer
+    : areActionsPaused
+      ? () => {}
+      : openDrawer
+
+  const iconColor = areActionsPaused ? '#ced5de' : '#2c405a'
+  return <div onClick={onClick} styleName={cx('reveal-actions-button', 'drawer', { 'drawer-close': !isDrawerOpen })} data-testid='reveal-actions-button'>
```
JettTech, line 398: Small request: Can we remove the ease-out property on the forward icon whenever `areActionsPaused` is true?  The fact that the opacity change on hover, makes the button feel like it is active/should trigger an action when pressed...

JettTech (changes requested):
One tiny ccs request and a small name change suggestion.  Otherwise looks good.  After updating this time, no need for a second request. :+1:

src/holofuel/pages/Inbox/Inbox.js
robbiecarlton: the issue is `isLoading` is good name for the state. That's what the state represents. I've used this convention in other projects when I wanted the setter function to do more than just set the state. Two possibilities:

1) we adopt this convention (the convention is `set[VariableName]Raw` is an acceptable name for a state setter, in addition to `set[VariableName]`, so whenever we see a setter that ends in `Raw` we know to just ignore that).

2) I rename the setter here back to `setIsLoading` and change the `setIsLoading` function later in this function to be called something else. `setIsLoadingAndAreActionsPaused` makes sense but is a mouthful. 

I have a slight preference for 1), but I'm used to it so if you feel strongly that you don't want to do that I'm happy to do 2)

src/holofuel/pages/Inbox/Inbox.js
robbiecarlton: Ah, good catch. I was testing on mobile sim so didn't see the hover behavior

src/holofuel/pages/Inbox/Inbox.js
JettTech: Got it.  I like the vibe of # 2, but agree that the name is a bit long.  This function's purpose is to pause the actionable slider for all other transactions on the page, so how about renaming the newer function to `SetIsPaused` or `PausePageActions`?  Thoughts?

robbiecarlton: well, the function is both to set paused and set loading. Set paused is dependent on set loading, so thats why I went with setLoading originally.

But I'll figure something out

robbiecarlton: Ok changed it to `setIsLoadingAndPaused`

How's that?

PR #260, opened 2020-05-05
Release v0.11.7

JettTech:
#### Fixed
- Detects hpos/conductor connection error using websocket timeouts.
#### Added
- Redirects to Login on connection error and displays user message.


JettTech:
@robbiecarlton @joel -
To Test:
In HC:
run `npm run start:holofuel-conductor-1` and `npm run start:live`
login and proceed to holofuel
shut down the conductor - after 3 ws connection timeouts, the app should redirect to the login page and display the connection error flash message

In Holo:
point the holo-nixpkgs to your local hp-admin
run hpos-shell | holo VM | holoport
login and proceed to holofuel
shut down hpos-shell | holo VM | holoport - after 3 ws connection timeouts, the app should redirect to the login page and display the connection error flash message

CHANGELOG.MD
```diff
@@ -5,6 +5,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## unreleased
 
+## [0.11.7] - 2020-04-30
```
robbiecarlton, line 8: :D This makes me so happy

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -22,32 +24,38 @@ export function PrimaryLayout ({
...
   }
 
-  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: 10000, onCompleted, onError, notifyOnNetworkStatusChange: true, ssr: false })
-  const { newMessage } = useFlashMessageContext()
+  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: 10000, onError, notifyOnNetworkStatusChange: true, ssr: false })
 
   useEffect(() => {
-    if (!isConnected) {
+    setIsConnected(isHposConnectionAlive && wsConnection)
+    if (isConnected === false) {
```
robbiecarlton: this is a little tricky, because you're setting `isConnected` and then immediately using the *old* value. In practice it's going to work here, but generally you should avoid this. Simplest way in this case would be by replacing

`if (isConnected === false) {`
with

`if (!(isHposConnectionAlive && wsConnection)) {`

Might be more readable to do

```
const newIsConnected = isHposConnectionAlive && wsConnection
setIsConnected(newIsConnected)
    if (!newIsConnected) {
```

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -22,32 +24,38 @@ export function PrimaryLayout ({
...
       newMessage('Your Holoport is currently unreachable.', 0)
+      if (window.location.pathname !== '/' && window.location.pathname !== '/admin/login') {
```
robbiecarlton: would be good to have this in a utility function in `utils/urls`, just so we can localize knowledge of what paths mean what. 
(I don't think we're doing anything like that very consistently yet, but we should be moving towards it)
-- this comment is not blocking.

src/holochainClient.js
```diff
@@ -212,6 +220,17 @@ export function createZomeCall (zomeCallPath, callOpts = {}) {
...
       return result
     } catch (error) {
+      // if ws timeout, redirect to login page and relay connection error
+      const timeout = /(timeout)/gi
+      const ws = /(ws)/gi
+      if (timeout.test(error) && ws.test(error)) {
+        wsTimeoutErrorVolume++
+        if (wsTimeoutErrorVolume >= 3) {
+          eraseHpAdminKeypair()
+          wsConnection = false
```
robbiecarlton: I think you want to reset `wsTimeoutErrorVolume` to zero here as well.

src/components/layout/PrimaryLayout/PrimaryLayout.js
robbiecarlton: Now wondering about moving the setting of `isConnected` into a separate wrapper component, like `LoadCurrentUser`. We already have it in context, so that would be simple enough, and it takes the responsibility away from `PrimaryLayout`, which makes sense to me.

-- this comment is not blocking.

robbiecarlton (changes requested):
@JettTech a couple of changes requested, and a couple of suggestions which I don't consider blocking (I've labeled those), but would be nice to have.

zo-el (approved):

PR #271, opened 2020-06-17
B-01229 - Hosting HoloFuel UI on Github Pages

robbiecarlton:
Had to do some slightly kludgy things to make it work with gh pages pretty inflexible hosting model. We will be able to remove some stuff once we're hosting on a normal service.

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## unreleased
+### Added
+- Github Pages deploy process
 ### Fixed
 - Missing style bug causing crash after accepting a transaction
 - Inbox Dont allow multiple transaction drawers to be open at once, but allow actioning a transaction while another is in process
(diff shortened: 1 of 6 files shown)


CHANGELOG.MD
```diff
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## unreleased
+### Added
+- Github Pages deploy process
```
JettTech, line 8: :tada:

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "hp-admin",
   "version": "0.11.11",
   "private": true,
+  "homepage": "https://holo-host.github.io/holofuel-static-ui/",
```
JettTech, line 5: Wouldn't it default to the repo name, ie `https://holo-host.github.io/hp-admin/`?  Are you able to name the url to any arbitrary string?

JettTech: Also, if we can update the name, can we just call it: `https://holo-host.github.io/holofuel-ui/` to match the app name itself?

JettTech (commented):
Nice work.  I just have one question/suggestion, but otherwise LGTM.

package.json
robbiecarlton: it's on its own repo, not `hp-admin`. We can't call it `holofuel-ui` because that repo is already taken.

Just a reminder that users will just see `holofuel.holo.host` as the url

PR #272, opened 2020-06-19
D-01068 - Accepted Requests appear as type request instead of as type offer

JettTech:
### Updates :
- corrects createOfferMutation returned transaction types
- handles balance rendering bug in sidebar (no longer flashes '--' following new poll to ledgerQuery)

## To Test : 
 - spin up two agents (ui and conductors) plus sim2h
 - create request from agent1 to agent2
 - have agent2 accept the request
 - notice that upon successful acceptance, the amount remains negative (ie: it is colored red and encased in the parentheses)


zo-el:
@JettTech what will this PR be addressing?

zo-el (approved):

JettTech:
@zo-el - I'll just document our discussion here for others' visibility.  This PR corrects the issue wherein after accepting a request, the request appears as an accepted offer rather than an accepted request.

PR #276, opened 2020-06-25
D-01067 - First render ws connection

JettTech:
### Updates : 
 - Ensure connection status monitoring upon the first load when the conductor is not yet started
 - Halt spinner whenever not connected

## To Test
 - start Agent 1 UI
 - notice connection failure message
 - start Agent 2 DNA
 - notice connection to Holochain conductor

PR #273, opened 2020-06-23
Accepted transactions optimistic rending

JettTech:
### Updates : 
 - ensures all non-inProcess actioned transactions remain hidden until the next successful fetch of list_pending
 - ensures all inProcess actioned transactions still render
 - coordinates the inProcess render timing to display the wording and highlighting at the same time (D-01069)
 - updates the inProcess highlighting to yellow
 - updates successful processing of inProcess transactions to render the green transaction highlight upon success

> For tickets : D-01069 & D-01072

## To Test
 - Spin up two Agents and sim2h
 - Create 1 offer from Agent 1 to Agent 2
 - Create 1 request from Agent 1 to Agent 2
 - Have Agent 2 accept the offer
 - Notice that inProcess rendering (updated highlight color and coordinated timing)
 - Notice that once the inProcess transaction successfully completed acceptance, the success message appears and the transaction is highlighted in green prior to disappearing
 - Have Agent 2 accept the request
 - Wait until the transaction is successful - notice that it never returns to inbox after clicking accept

PR #281, opened 2020-06-29
Release v0.12.X updates

JettTech:
### Updates : 
- PR #268 : 
  - reset the redirect path to default to home path
- PR #269 :
  - manages new event state in actionable promises, as provided from DNA
  - adds counter and functionality to trigger accepting of incomplete invoiced offers
  - adds user message to display if timeout error occurs upon accepting an offer
- PR #270 :
  - uses request options from useQuery to dynamically stop/start polling
- PR #272 :
  - corrects createOfferMutation returned transaction types
  - handles balance rendering bug in sidebar (no longer flashes '--' following new poll to ledgerQuery)
- PR #273 :
  - ensures all non-inProcess actioned transactions remain hidden until the next successful fetch of list_pending
  - ensures all inProcess actioned transactions still render
  - coordinates the inProcess render timing to display the wording and highlighting at the same time (D-01069)
  - updates the inProcess highlighting to yellow
  - updates successful processing of inProcess transactions to render the green transaction highlight upon success
- PR #274 : 
  - removes redundant polling/network calls
- PR #276, #278 , #279  :
  - ensure connection status monitoring upon the first load when the conductor is not yet started (handled in the updated version of hc-web-client)
  - update hc-web-client to v0.5.3
  - halt spinner whenever not connected
  - update all HPOS env to start default ws connection to true
- PR #277 :
  -  ensures that only pending payments (ie: accepted requests) are accepted automatically
- PR #280 :
  - decreases polling interval to 60000ms
  - adds refresh button in Side Navigation Bar


src/components/HashIcon/HashIcon.js
```diff
@@ -1,19 +1,30 @@
...
 import React from 'react'
+import cx from 'classnames'
 import { Network as Identicon } from 'react-identicon-variety-pack'
 import './HashIcon.module.css'
 
+function DefaultIdenticon ({
+  size,
+  className
+}) {
+  return <div styleName={cx({ 'big-icon': size > 40 }, { 'small-icon': size < 40 })} className={className} />
```
robbiecarlton: Would be better to just pass size in to a `style` prop, rather than just having two possible fixed sizes. This is how the actual `Identicon` component works, so it makes sense for this to be the same

```
<div styleName='default-identicon' style={{ width: size, height: size }} className={className} />
```

robbiecarlton: The current solution will break if we add a third size identicon anywhere

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -123,7 +127,25 @@ function presentPendingRequest (transaction, annuled = false) {
...
 
-function presentPendingOffer (transaction, annuled = false) {
+let counter = 0
+function presentPendingOffer (transaction, invoicedOffers = [], annuled = false) {
+  const invalidEvent = invoicedOffers.find(io => !io.Invoice)
+  if (invalidEvent) return new Error(`Error: invalidEvent found: ${invalidEvent}.`)
+  const handleEvent = () => HoloFuelDnaInterface.offers.accept(transaction.event[0])
+  const findEvent = () => {
+    const invoice = invoicedOffers.find(io => io.Invoice)
+    console.log('---- >>', counter)
```

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -27,6 +28,11 @@ import { presentAgentId, presentHolofuelAmount, sliceHash, useLoadingFirstTime,
 import { caribbeanGreen } from 'utils/colors'
 import { OFFER_REQUEST_PATH } from 'holofuel/utils/urls'
 import { TYPE, STATUS, DIRECTION, shouldShowTransactionInInbox } from 'models/Transaction'
+import { userNotification } from 'data-interfaces/HoloFuelDnaInterface'
```
robbiecarlton: we should avoid this pattern. I'm surprised it works. 

Instead we should do something like have a `stale` flag on the transaction object, and catch that in the frontend.

But let's talk about exactly what the problem is that this solves and figure out a more robust way of doing it.

src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -362,7 +407,11 @@ export function TransactionRow ({ transaction, setConfirmationModalProperties, i
       {/* {isActionable ? <div /> : <div styleName='balance'>{presentBalance}</div>} */}
     </div>
 
-    {isLoading && <Loading styleName='transaction-row-loading' width={20} height={20} />}
+    {isLoading && !inProcess && <Loading styleName='transaction-row-loading' width={20} height={20} />}
+
+    {inProcess && !highlightGreen && <ToolTip toolTipText={timeoutErrorMessage}>
+      <h4 styleName='alert-msg'>{offerInProcessMessage}</h4>
```
robbiecarlton: Can we just have this value be inline?
```
 <h4 styleName='alert-msg'>acceptance pending</h4>
```

src/graphql/HolofuelActionableTransactionsQuery.gql
```diff
@@ -11,5 +11,7 @@ query HolofuelActionableTransactionsQuery {
     fees
     isPayingARequest
     canceledBy { id, nickname }
+    inProcess
+    actioned
```
robbiecarlton: do we definitley need separate `inProcess` and `actioned` fields? Aren't they negations of each other? Or have a similar relationship that makes having both redundant? Let's talk about this, I might be missing something.

src/holofuel/components/HashIcon/HashIcon.js
```diff
@@ -1,19 +1,30 @@
...
 import React from 'react'
+import cx from 'classnames'
 import { Network as Identicon } from 'react-identicon-variety-pack'
 import './HashIcon.module.css'
 
+function DefaultIdenticon ({
+  size,
+  className
+}) {
+  return <div styleName={cx({ 'big-icon': size > 40 }, { 'small-icon': size < 40 })} className={className} />
```
robbiecarlton: see comments on the other copy of `HashIcon`

src/holofuel/components/SideMenu/SideMenu.js
```diff
@@ -63,11 +67,19 @@ export default function SideMenu ({
               Profile
             </Link>
           </li>
-          {process.env.REACT_APP_HOLOFUEL_APP !== 'true' && <li>
-            <Link to='/admin/' styleName='nav-link'>
+          {process.env.REACT_APP_HOLOFUEL_APP !== 'true' && <li styleName='last-list-item'>
+            <Link to='/admin/' styleName='nav-link last-nav-link'>
```
robbiecarlton: can you do this using [:last-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child) instead of a separate style?

src/holofuel/pages/Home/Home.js
```diff
@@ -45,7 +55,7 @@ export default function Home () {
   const history = useHistory()
   const goToOfferRequest = () => history.push(OFFER_REQUEST_PATH)
 
-  const isLoadingFirstPendingTransactions = useLoadingFirstTime(loadingTransactions)
+  const isLoadingFirstPendingTransactions = useLoadingFirstTime(isLoadingTransactions)
```
robbiecarlton: I don't think you need the `useEffect` above. Instead, this line can just be

```
const isLoadingFirstPendingTransactions = useLoadingFirstTime(isConnected && loadingTransactions)
```

src/holofuel/pages/Inbox/Inbox.test.js
```diff
@@ -77,13 +79,15 @@ const actionableTransactions = pendingList.requests.concat(pendingList.promises)
...
 describe('Inbox connected (with Agent Nicknames)', () => {
-  it('renders', async () => {
+  it.skip('renders', async () => {
     const { getAllByRole, getByText } = await renderAndWait(<ApolloProvider client={apolloClient}>
       <Inbox history={{}} />
     </ApolloProvider>, 1500)
 
     expect(getByText(`${presentHolofuelAmount(ledger.balance)} TF`)).toBeInTheDocument()
 
+    console.log('actionableTransactions : ', actionableTransactions)
```
robbiecarlton: remove log

src/holofuel/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -19,20 +23,30 @@ import { wsConnection } from 'holochainClient'
...
+  const { loading: actionableTransactionsLoading, data: { holofuelActionableTransactions: actionableTransactions = [] } = {}, refetch: refetchActionableTransactions, stopPolling: stopPollingActionableTransactions, startPolling: startPollingActionableTransactions } = useQuery(HolofuelActionableTransactionsQuery, { fetchPolicy: 'cache-and-network' })
+  const { loading: completedTransactionsLoading, refetch: refetchCompletedTransactions, stopPolling: stopPollingCompletedTransactions, startPolling: startPollingCompletedTransactions } = useQuery(HolofuelCompletedTransactionsQuery, { fetchPolicy: 'cache-and-network' })
+  const { loading: nonPendingTransactionsLoading, refetch: refetchNonPendingTransactions } = useQuery(HolofuelNonPendingTransactionsQuery, { fetchPolicy: 'cache-and-network' })
+  const { loading: waitingTransactionsLoading, refetch: refetchWaitingTransactions } = useQuery(HolofuelWaitingTransactionsQuery, { fetchPolicy: 'cache-and-network' })
```
robbiecarlton: I think it might be tidier to wrap all of these in a `useTransactionQueries` hook, and just return a couple of things from that

I think all you need is

refetchCalls, and [stopPolling, startPolling] X [ActionableTransactions, CompletedTransactions]

robbiecarlton (changes requested):
Some changes, including a couple of things I'd prefer to discuss synchronously if possible. 

Also, I noticed a few skipped tests. Was that intentional? What's the plan with those? Just skipping them so you can move fast for the release process?

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -35,25 +35,46 @@ export function PrimaryLayout ({
...
-      newMessage('Your Holoport is currently unreachable.', 0)
+    if (!isConnected.hpos) {
+      newMessage('Your Holoport is currently unreachable. Attempting to reconnect.', 0)
+      // reroute to login on network/hpos connection error
       if (window.location.pathname !== '/' && window.location.pathname !== '/admin/login') {
         push('/')
       }
     }
+
+    if (window.location.pathname !== '/' && window.location.pathname !== '/admin/login') {
```
robbiecarlton: This is repeated enough that it would be good to move this test into a util function.

```
function isLoginPage(window) {
  return window.location.pathname !== '/' && window.location.pathname !== '/admin/login'
}
```

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -35,25 +35,46 @@ export function PrimaryLayout ({
...
+      // if inside happ, check for connection to holochain
+      if (isConnected.hpos && !isConnected.holochain) {
+        newMessage('Your Holochain Conductor is currently unreachable.', 0)
+        // reroute to dashboard on ws connection / hc conductor failure
+        if (window.location.pathname !== '/admin' && window.location.pathname !== '/admin/' && window.location.pathname !== '/admin/dashboard') {
+          push('/admin/dashboard')
```
robbiecarlton: why reroute to dashboard here?

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -35,25 +35,46 @@ export function PrimaryLayout ({
     setIsHposConnectionAlive(isHposConnectionActive)
   }
 
-  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: 10000, onError, notifyOnNetworkStatusChange: true, ssr: false })
+  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: 30000, onError, notifyOnNetworkStatusChange: true, ssr: false })
 
   useInterval(() => {
-    setIsConnected(isHposConnectionAlive && wsConnection)
+    if (window.location.pathname === '/' || window.location.pathname === '/admin/login') {
```
robbiecarlton: I think all of this connection management logic should probably be happening in a separate wrapper component near the top of the component tree, rather than in PrimaryLayout

src/contexts/useConnectionContext.js
```diff
@@ -3,7 +3,10 @@ import React, { createContext, useContext, useState } from 'react'
 export const ConnectionContext = createContext()
 
 export function ConnectionProvider ({ children }) {
-  const [isConnected, setIsConnected] = useState(false)
+  const [isConnected, setIsConnected] = useState({
```
robbiecarlton: should rename this variable now. `isConnected` tells me it's a boolean. `connectionStatus` is probably good

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -123,7 +126,23 @@ function presentPendingRequest (transaction, annuled = false) {
   return presentRequest({ origin, event: event[2], stateDirection, status, type, eventTimestamp, counterpartyId, amount, notes, fees: fee })
 }
 
-function presentPendingOffer (transaction, annuled = false) {
+let counter = 0
+function presentPendingOffer (transaction, invoicedOffers = [], annuled = false) {
+  const invalidEvent = invoicedOffers.find(io => !io.Invoice)
+  if (invalidEvent) return new Error(`Error: invalidEvent found: ${invalidEvent}.`)
```
robbiecarlton, line 131: You should just throw the error here. You don't want to return it, that will just give you a different and more confusing error up the chain when the consumer tries to treat the returned error as a presentedOffer.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -123,7 +126,23 @@ function presentPendingRequest (transaction, annuled = false) {
...
+  const handleEvent = () => HoloFuelDnaInterface.offers.accept(transaction.event[0])
+  const findEvent = () => {
+    const invoice = invoicedOffers.find(io => io.Invoice)
+    counter++
+    if (invoice) {
+      if (counter > 10) {
+        counter = 0
+        handleEvent()
```
robbiecarlton: I'm not entirely sure what this is for, but we almost certainly don't want it in this `present` function. This function should have no side effects and just turn a holofuel dna shaped transaction into a ui shaped transaction

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -394,34 +431,86 @@ const HoloFuelDnaInterface = {
     create: async (counterpartyId, amount, notes, requestId) => {
       const origin = await createZomeCall('transactions/promise')(pickBy(i => i, { to: counterpartyId, amount: amount.toString(), deadline: mockDeadline(), notes, request: requestId }))
 
-      return {
-        id: requestId || origin, // NB: If requestId isn't defined, then offer use origin as the ID (ie. Offer is the initiating transaction).
+      const presentableTransaction = {
```
robbiecarlton: minor quibble, but I would call this `presentedTransaction`

src/components/AlphaFlag/AlphaFlag.module.css
```diff
@@ -3,7 +3,7 @@
 }
 
 .flag {
-  background-color: var(--color-havelock-blue);
+  background-color: #6D5AAD;
```
robbiecarlton: can we have a color variable here

src/components/SideMenu/SideMenu.module.css
```diff
@@ -28,7 +28,7 @@
 .drawer--open .container {
   transform: translateX(100%);
   pointer-events: initial;
-  overflow-y: scroll;
+  /* overflow-y: scroll; */
```
robbiecarlton: remove comment

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -14,67 +13,126 @@ import useFlashMessageContext from 'contexts/useFlashMessageContext'
...
+  const [isInsideApp, setIsInsideApp] = useState(true)
   const [isHposConnectionAlive, setIsHposConnectionAlive] = useState(true)
   const { setIsConnected, isConnected } = useConnectionContext()
   const { setCurrentUser } = useCurrentUserContext()
   const { newMessage } = useFlashMessageContext()
   const { push } = useHistory()
 
+  console.log('useConnectionContext : ', useConnectionContext);
+  console.log('isConnected : ', isConnected);
+  console.log('setIsConnected : ', setIsConnected);
```
robbiecarlton: remove logs

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -14,67 +13,126 @@ import useFlashMessageContext from 'contexts/useFlashMessageContext'
...
-    setIsConnected(isHposConnectionAlive && wsConnection)
+    if (isInsideApp) {
+      setIsConnected({ hpos: isHposConnectionAlive, holochain: wsConnection })
+    }
   }, 5000)
 
   useEffect(() => {
-    if (isConnected) {
-      newMessage('', 0)
+    setIsInsideApp(window.location.pathname !== ROOT && window.location.pathname !== HP_ADMIN_LOGIN)
```
robbiecarlton: I'd like all this connection logic to be in it's own wrapper component

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -133,7 +143,30 @@ function presentPendingOffer (transaction, annuled = false) {
...
+    promises.forEach(p => {
+      if (!isEmpty(p[1])) {
+            findEvent(p[0],p[1])
+      }
+    })
+  }
+  return {requests, promises, declined }
+}
+
+const findEvent = async (tx, invoicedOffers) => {
```
robbiecarlton: this isn't finding an event, what is it doing? Let's call it that.

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -133,7 +143,30 @@ function presentPendingOffer (transaction, annuled = false) {
...
+  const { requests, promises, declined } = await createZomeCall('transactions/list_pending')(params)
+  counter++
+  if (counter === 8) {
+    counter = 0
+    promises.forEach(p => {
+      if (!isEmpty(p[1])) {
+            findEvent(p[0],p[1])
+      }
+    })
+  }
```
robbiecarlton: This code has side effects. The DNAInterfaces should as much as possible just be pure data transformations, and not have side effects. I would move this logic into the UI

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -295,8 +337,8 @@ const HoloFuelDnaInterface = {
       return presentedNonActionableTransactions.sort((a, b) => a.timestamp > b.timestamp ? -1 : 1)
     },
     getPending: async (transactionId) => {
-      const { requests, promises } = await createZomeCall('transactions/list_pending')({ origins: transactionId })
-      const transactions = requests.map(r => presentPendingRequest(r)).concat(promises.map(p => presentPendingOffer(p)))
+      const { requests, promises } = await getListPending({ origins: transactionId })
+      const transactions = await requests.map(r => presentPendingRequest(r)).concat(promises.map(p => presentPendingOffer(p[0], p[1]))).filter(tx => !(tx instanceof Error))
```
robbiecarlton: seems like we don't need this await

src/graphql-server/schema.graphql
```diff
@@ -68,8 +68,11 @@ type Transaction {
   presentBalance: Float
   notes: String
   happName: String
-  isPayingARequest: Boolean
   canceledBy: HolofuelUser
+  isPayingARequest: Boolean
+  inProcess: Boolean
+  actioned: Boolean
+  stale: Boolean
```
robbiecarlton: I would call all these booleans `isSomething`, to make it clear they're booleans

src/holofuel/utils/urls.js
```diff
@@ -1,6 +1,6 @@
-export const HOME_PATH = 'home'
+// export const HOME_PATH = 'home'
 export const INBOX_PATH = 'inbox'
 export const OFFER_REQUEST_PATH = 'offer-request'
 export const HISTORY_PATH = 'history'
 export const PROFILE_PATH = 'profile'
-export const HP_ADMIN_LOGIN_PATH = '/admin/login'
+export const HP_ADMIN_DASHBOARD = '/admin/dashboard'
```
robbiecarlton: call this `HP_ADMIN_DASHBOARD_PATH`

src/pages/Dashboard/Dashboard.js
```diff
@@ -18,16 +18,26 @@ export const mockEarnings = 4984
...
-  const { data: { holofuelLedger: { balance } = { balance: 0 } } = {} } = useQuery(HolofuelLedgerQuery)
-
   const isEarningsZero = Number(earnings) === 0
   const isBalanceZero = Number(balance) === 0
 
   const greeting = !isEmpty(settings.hostName) ? `Hi ${settings.hostName}!` : 'Hi!'
+  // const [urlOrigin, setUrlOrigin] = useState()
+  const location = useLocation()
+
+  useEffect(() => {
```
robbiecarlton: remove this stuff

robbiecarlton (approved):
Some small changes, and some larger, do what you can and then it's good to merge

src/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: This is was re-evaluated and is already updated in the update additions PR.

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Good catch; this is definitely residue from prior versions of the code.

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: It's being returned here so that is can be caught in the error filter (ie: `filter(tx => !(tx instanceof Error))`)  - see both the `allActionable` query (on line 287) and the `getPending` query (on line 343).  

The intent of this approach is to ensure that if one transaction errors, the remainder of the transactions are still able to be returned and rendered in the UI.

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: Let's talk about this more.

PR #282, opened 2020-06-29
D-01084 & D-01083 - 2020 06 29/login and counterparty fix

JettTech:
### Updates : 
 - Separates holofuel and hpos connection monitoring in hp-admin app/hpos environment
   - if holofuel conductor is down, the app will redirect to the dashboard and use a flash message to signal that the conductor is down
   - if hpos connection is down, the app will redirect to the login page and use a flash message to signal that the conductor is down
   - if hpos connection is up, then the user can log into the app, regardless of whether the ws connection the conductor is up
- Returns a default value for counterparty when an error is returned from `get_profile` zome call. 

###### Test the holochain conductor connection
 - Spin up local agent 1 in hpos env and sim2h server (ie: `npm run start:live` and `npm run start:holofuel-conductor-1`)
 - Sign into app and navigate to holofuel 
 - Shut down the conductor
 - Notice the page redirect to the hp-admin dashboard and the flash message appear with the message about the connection to the conductor
###### Test the counterparty
- Turn back on the conductor and spin up agent2
- Create and complete a new transaction between agent1 and agent2
- Turn off the sim2h
- Navigate each agent to their history page and notice the transaction still appearing, showing the counterparty name, if cached, or by the last 6 of their agent id, if not.

> HPOS
###### Test the hpos connection
- Point your holo-nixpkgs to your local version of hp-admin (make sure it is on the `2020-06-29/login-and-counterparty` branch)
- Enter into your hpos-shell
- Open hp-admin at your correct agent url and sign in
- Exit your hpos-shell
- Notice the page redirect to the login page and the flash message appear with the message about the hpos connection
- Enter back into hpos-shell
- Log back into app (without having to refresh)


zo-el (changes requested):
We need to fix 
- receive erroneous 'incorrect password' message when connection is down and trying to log in
- flash message alert for networking error is finicky

src/holochainClient.js
```diff
@@ -114,14 +114,14 @@ let holochainClient
...
 export let wsConnection = holochainClient
   ? true
   : process.env.NODE_ENV === 'test'
     ? true
-    : process.env.REACT_APP_HOLOFUEL_APP === 'true'
-      ? false
-      : true
+    : !process.env.REACT_APP_HOLOFUEL_APP
+
+console.log('wsConnection : ', wsConnection)
```
zo-el, line 124: we can remove this

zo-el (approved):
After being redirected to the login page the UI thinks it is connected for sometime till it is proven otherwise. 

This is not a good user experience. 

Lets make a new ticket for this and move ahead with the next round of testing for now.

Also I have see a bunch of times that the zome calls not get timed out which only leaves the user to do a hard refresh. We should look into this issue too.

PR #283, opened 2020-06-29
D 01071 - 2020-06-29/null notes value

JettTech:
### Updates : 
 - Only displays notes field if notes field exists, fix in Activity Tab of Inbox page for declined transactions

## To Test : 
 - Spin up two agents and sim2h
 - Have agent1 offer an amount to agent2 - ensuring there is no text in the notes field.
 - Have agent2 decline the offer
 - Visit the Inbox Activity view and notice that the transaction notes displays an empty string instead of 'null'.


src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -35,24 +35,38 @@ export function PrimaryLayout ({
...
   useEffect(() => {
-    if (isConnected) {
-      newMessage('', 0)
-      setCurrentUser({
-        hostPubKey: settings.hostPubKey,
-        hostName: settings.hostName || ''
-      })
-    } else {
+    if (!isConnected.hpos) {
+      console.log('isConnected.hpos (should be false) : ', isConnected.hpos)
```
zo-el: remove logs

zo-el (approved):

PR #285, opened 2020-06-30
D 01089 - Receive payents pending error

JettTech:
### Updates
 - adds try-catch to handle internal errors that can't be parsed, such as the receive_payments_pnding error


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -447,18 +450,25 @@ const HoloFuelDnaInterface = {
...
-            return {
-              ...transaction,
-              id: transactionId, // should always match `Object.entries(result)[0][0]`
-              direction: DIRECTION.incoming, // this indicates the hf recipient
-              status: STATUS.pending,
-              type: TYPE.offer,
-              actioned: true
+          } else {
+            try {
+              const parseableResponse = JSON.parse(acceptedPaymentHash.Err.Internal)
```
zo-el: It will fail here itself so. we do not need to check if `acceptedPaymentHash.Err.Internal` can be parsed.

Just the try catch will do the trick

zo-el (changes requested):

src/data-interfaces/HoloFuelDnaInterface.js
JettTech: I did this to make it more legible, but I'm fine with having it just check/fail on the if statement instead. 👍

zo-el (approved):

zo-el (approved):

PR #288, opened 2020-07-03
B-01172 hosting

robbiecarlton:


JettTech (approved):
LGTM. One note, it does look like we are missing a link back to the dashboard from settings. This seems to be also missing from the design, so I won't mark it as a blocker, but adding it here to follow-up.

robbiecarlton:
> LGTM. One note, it does look like we are missing a link back to the dashboard from settings. This seems to be also missing from the design, so I won't mark it as a blocker, but adding it here to follow-up.

@JettTech - yep, thats being handled in a separate ticket

PR #291, opened 2020-07-08
B-01900 - ui feedback edits

JettTech:
### Updates:
UI/UX Edits as follows :
 #### hp-admin:
 - replace avatar hash with a gear icon
 - have holofuel open up into different tab
 - when on settings page, add a back arrow icon to return to the dashboard page
 - remove the side navbar
 - create a banner for alpha notice and legal info/links

 #### holofuel:
 - implement a responsive layout with desktop design pattern
 - indicate current/active page in the side navbar
 - update title in the header
 - remove the balance from side nav
 - add back arrow in front of hp admin to increase visual/cognitive distinction between returning to hp admin happ and navigating between holofuel pages

### DIFF
--- a/src/holofuel/components/CopyToClipboard/CopyToClipboard.js
+++ b/src/holofuel/components/CopyToClipboard/CopyToClipboard.js
@@ -13,7 +13,7 @@ export default function CopyToClipboard ({
 
   const handleCopyItem = async () => {
     const wasCopied = await copy(copyContent)
-    if (wasCopied === true) newMessage(messageText, 5000)
+    if (wasCopied === true) newMessage(messageText,  0)
   }
 
   return <div onClick={handleCopyItem} styleName='copy-item'>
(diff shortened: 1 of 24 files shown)


src/holofuel/components/CopyToClipboard/CopyToClipboard.js
```diff
@@ -13,7 +13,7 @@ export default function CopyToClipboard ({
 
   const handleCopyItem = async () => {
     const wasCopied = await copy(copyContent)
-    if (wasCopied === true) newMessage(messageText, 5000)
+    if (wasCopied === true) newMessage(messageText,  0)
```
zo-el: Why is this timeout changed?

zo-el (changes requested):
Testing failed. I was not able to get to the holofuel app from hp-admin. It kept asking for the password and then opened hp-admin again.

src/holofuel/components/CopyToClipboard/CopyToClipboard.js
JettTech: Good catch; this is def residue from testing... changing it back now.

zo-el:
```suggestion
    if (wasCopied === true) newMessage(messageText,  5000)
```

PR #292, opened 2020-07-10
Update memory leak

zo-el:


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -320,8 +333,8 @@ const HoloFuelDnaInterface = {
       return presentedNonActionableTransactions.sort((a, b) => a.timestamp > b.timestamp ? -1 : 1)
     },
     getPending: async (transactionId) => {
-      const { requests, promises } = await createZomeCall('transactions/list_pending')({ origins: transactionId })
-      const transactions = requests.map(r => presentPendingRequest(r)).concat(promises.map(p => presentPendingOffer(p[0], p[1]))).filter(tx => !(tx instanceof Error))
+      const { requests, promises } = await getListPending()
```
JettTech: This is missing the transactionId and will not return the correct data.

JettTech:
```suggestion
      const { requests, promises } = await getListPending({ transactionId })
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -159,6 +148,30 @@ function presentPendingOffer (transaction, invoicedOffers = [], annuled = false)
   return presentOffer({ origin, event: event[2], stateDirection, status, type, eventTimestamp, counterpartyId, amount, notes, fees: fee, isPayingARequest, inProcess })
 }
 
+let counter = 0
+async function getListPending () {
```
JettTech:
```suggestion
async function getListPending ({ transactionId }) {
```

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -159,6 +148,30 @@ function presentPendingOffer (transaction, invoicedOffers = [], annuled = false)
...
+  const { requests, promises, declined } = await createZomeCall('transactions/list_pending')()
```
JettTech:
```suggestion
  const params = transactionId ? { transactionId } : {}
  const { requests, promises, declined } = await createZomeCall('transactions/list_pending')(params)
```

JettTech (changes requested):
Update the `getPending` call to include the params; otherwise, this is good to go. :+1:

PR #293, opened 2020-07-10
B-01917 - polling interval global + modal width update

JettTech:
- adds a command-line arg for the `start:live` script which accepts a number to set for the polling amount
- updates modal in desktop settings to match updated width


src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -34,7 +34,7 @@ export function PrimaryLayout ({
     setIsHposConnectionAlive(isHposConnectionActive)
   }
 
-  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: 10000, onError, notifyOnNetworkStatusChange: true, ssr: false })
+  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: (POLLING_INTERVAL / 3), onError, notifyOnNetworkStatusChange: true, ssr: false })
```
robbiecarlton: let's export a separate Constant called `POLLING_INTERVAL_SETTINGS`

src/utils/index.js
```diff
@@ -2,6 +2,9 @@ import { useState, useEffect, useRef } from 'react'
 import { flow, groupBy, keys, sortBy, reverse } from 'lodash/fp'
 import moment from 'moment'
 
+// Default polling interval to 30000ms
+export const POLLING_INTERVAL = parseInt(process.env.REACT_APP_POLLING_INTERVAL) || 30000
+
```
robbiecarlton: add 
```
export const POLLING_INTERVAL_SETTINGS = Math.max(Math.floor(POLLING_INTERVAL), 1000)
```

and add a similar math.max constraint to `POLLING_INTERVAL` above

robbiecarlton (changes requested):
Small change and then it's good to merge

src/utils/index.js
```diff
@@ -2,6 +2,12 @@ import { useState, useEffect, useRef } from 'react'
 import { flow, groupBy, keys, sortBy, reverse } from 'lodash/fp'
 import moment from 'moment'
 
+// Default polling interval to 30000ms
+export const POLLING_INTERVAL = parseInt(process.env.REACT_APP_POLLING_INTERVAL) || 30000
```
robbiecarlton: we don't need to default it to 30000 here *and* in the scripts. I suggest just doing it here and get rid of the default in the scripts.

src/utils/index.js
```diff
@@ -2,6 +2,12 @@ import { useState, useEffect, useRef } from 'react'
...
+export const POLLING_INTERVAL_GENERAL = Math.max(Math.floor(POLLING_INTERVAL), 5000)
+// settings is called x3 as frequently for hpos connection monitoring, but the min should still be set to 5000ms (ie: 15000/3)
+export const POLLING_INTERVAL_SETTINGS = Math.max(Math.floor(POLLING_INTERVAL), 15000)
```
robbiecarlton: shouldn't this be

```
export const POLLING_INTERVAL_SETTINGS = Math.max(Math.floor(POLLING_INTERVAL / 3), 5000)
```

and remove the `/ 3` in the actual settings page.

The point of having this separate const was to set it to the value we want here.

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -9,85 +9,106 @@ import Header from 'components/Header'
...
 
+  const [isPausedConnectionCheckInterval, setIsPausedConnectionCheckInterval] = useState(false)
+  const [userMessage, setUserMessage] = useState('')
+
   const onError = ({ graphQLErrors: { isHposConnectionActive } }) => {
     setIsHposConnectionAlive(isHposConnectionActive)
   }
 
-  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: 10000, onError, notifyOnNetworkStatusChange: true, ssr: false })
+  const { data: { hposSettings: settings = {} } = {} } = useQuery(HposSettingsQuery, { pollInterval: (POLLING_INTERVAL_SETTINGS / 3), onError, notifyOnNetworkStatusChange: true, ssr: false })
```
robbiecarlton: see comment on `utils/index.js`

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -9,85 +9,106 @@ import Header from 'components/Header'
...
   const [isHposConnectionAlive, setIsHposConnectionAlive] = useState(true)
-  const { setConnectionStatus, connectionStatus } = useConnectionContext()
+  const { connectionStatus, setConnectionStatus } = useConnectionContext()
+  const { isConnected: isHFConductorConnected } = useHFConnectionContext()
   const { setCurrentUser } = useCurrentUserContext()
   const { newMessage } = useFlashMessageContext()
   const { push } = useHistory()
 
+  const [isPausedConnectionCheckInterval, setIsPausedConnectionCheckInterval] = useState(false)
+  const [userMessage, setUserMessage] = useState('')
```
robbiecarlton: I'm not seeing where the user is ever seeing the value of this state, so calling it `userMessage` makes the code harder to read and maintain. AFAICT, what this state is doing is just tracking which of 3 states the app is in, so I might just call it `connectionStatus`, and use constants to populate it.

```
const [connectionStatus, setConnectionStatus] = useState(CONNECTION_STATUS.connected)
```

and then eg on line 75

```
setConnectionStatus(CONNECTION_STATUS.noHoloport)
```

I think this makes it much clearer what's going on.

scripts/build-holofuel.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Default polling interval to 30000ms
+POLLING_INTERVAL=${1:-'30000'}
```
robbiecarlton: see comment in `utils/index.js`

robbiecarlton (changes requested):

src/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: The userMessage is the message that is displayed back to the user in a flash message.  On line 76, 87, and 93, the `userMessage` is set to `newMessage` to render a flash message with the current message.

robbiecarlton: right, but there's no reason to keep that in state.

You're not going

`setNewMessage(userMessage)`

you're going

```
setUserMessage(someConstant)
setNewMessage(someConstant)
```

The fact that you're storing the same constant in `userMessage` as you happen to be displaying to the user is irrelevant and confusing.

scripts/build-holofuel.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-# Default polling interval to 30000ms
-POLLING_INTERVAL=${1:-'30000'}
+# Set polling var if passed in
+# (polling default is set in app utils)
+POLLING_INTERVAL=${1:-'0'}
```
robbiecarlton: setting this to a default of '0' is fragile and a little misleading. Can we not just have no default so that line 14 of this file reads

```
REACT_APP_VERSION=$npm_package_version REACT_APP_HOLOFUEL_APP=true REACT_APP_RAW_HOLOCHAIN=true REACT_APP_POLLING_INTERVAL=$1 node scripts/build.js
```

Does that break? If not we should do that.

robbiecarlton (changes requested):

scripts/build-holofuel.sh
JettTech: Well, in the scripts we're doing a validation check to make sure the variable passed in, is indeed a number... such that the app won't break.  For this same reason, I'm adding a default `0` in the case where none is provided, which is later set to the standard app default of 3000ms in utils.

I do think it is important to have a number validation of the arg passed in.... but that does also require us to set a default number value, instead of nothing, or null...

src/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: Got it... ok, I can see how to make this clearer.  Updates soon to come. :+1:

scripts/build-holofuel.sh
JettTech: Do you prefer the default to 3000ms?

robbiecarlton: How I'd do this is I would only validate that the command line arg is a number if it's passed in. If no arg is passed in, skip the validation.

But this isn't a big deal so I don't want to block. Maybe just add a comment saying "this default of 0 is overridden in the app"

robbiecarlton (approved):

robbiecarlton (approved):

PR #304, opened 2020-07-17
D-01105 - correct alpha banner display for mobile + small edits

JettTech:
### Updates:
 - converts alpha banner into a non-sticky footer for mobile view
 - manages NaN state when rendering HF/TF amounts
 - corrects isLoginPage naming to logic
 - improves connection feedback rendering

### DIFF
--- a/src/components/layout/PrimaryLayout/PrimaryLayout.js
+++ b/src/components/layout/PrimaryLayout/PrimaryLayout.js
@@ -117,7 +117,7 @@ export function PrimaryLayout ({
       {showHeader && <Header
         {...headerProps}
         settings={connectionStatus.hpos ? settings : {}} />}
-      {showAlphaFlag && <AlphaFlag styleName='styles.alpha-flag' />}
+      {showAlphaFlag && <AlphaFlag styleName='styles.alpha-flag-page' />}
       <div styleName='styles.content'>
         <FlashMessage />
         {children}
@@ -128,7 +128,7 @@ export function PrimaryLayout ({
       <div styleName={cx('styles.container', { 'styles.container-wide': isWide })}>
         <footer styleName='styles.footer'>
           <div styleName='styles.alpha-info'>
-            <AlphaFlag variant='right' styleName='styles.alpha-flag' />
+            <AlphaFlag variant='right' styleName='styles.alpha-flag-banner' />
             <p>
               HP Admin is in Alpha testing.
             </p>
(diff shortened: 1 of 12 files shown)


src/utils/urls.js
```diff
@@ -5,5 +5,5 @@ export const HP_ADMIN_DASHBOARD_PATH = '/admin/dashboard'
 export const HP_ADMIN_SETTINGS_PATH = '/admin/settings'
 
 export function isLoginPage (window) {
-  return window.location.pathname !== '/' && window.location.pathname !== '/admin/login'
+  return window.location.pathname === '/' || window.location.pathname === '/admin/login'
```
robbiecarlton, line 8: 👍

src/components/layout/PrimaryLayout/PrimaryLayout.js
```diff
@@ -26,68 +27,88 @@ export function PrimaryLayout ({
 }) {
   const [isInsideApp, setIsInsideApp] = useState(true)
   const [isHposConnectionAlive, setIsHposConnectionAlive] = useState(true)
-  const { setConnectionStatus, connectionStatus } = useConnectionContext()
+  const { connectionStatus, setConnectionStatus } = useConnectionContext()
+  const { isConnected: isHFConductorConnected } = useHFConnectionContext()
```
robbiecarlton, line 31: why do we need HFConnectionContext in hp-admin primarylayout?

I would rather nothing in the `hp-admin` side of the tree import anything from the `/holofuel` if at all possible.

What problem does this solve?

robbiecarlton (changes requested):

src/components/layout/PrimaryLayout/PrimaryLayout.js
JettTech: This solves the speed of communication regarding connectivity.  

The scenario is, when the conductor goes down while transacting on holofuel, the holofuel app is the first to know and trigger any polling updates or rerouting logic.  

Whenever in the holo case, we inform the app to reroute to the login page (as a disconnection to the conductor is quickly followed by a disconnection to hpos), however as the login page is apart of the hp-admin side of the app, and it's connection context last remembers the connection as active/true, then it takes another few seconds for the login page to realize the connection is down while it renders.  

In sum, the user ends up being logged out, but sits there for a few seconds before having any feedback about why.... unless we can use the context from the holofuel app to inform us earlier.

robbiecarlton: hmm. We're going to need a new plan once we're on Holofuel 1.x

robbiecarlton (approved):

PR #305, opened 2020-07-20
B 01916 - tx obj with counterparty name

JettTech:
### Updates:
 - updates UI with new pattern for returning counterparty nickname


src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -223,30 +232,18 @@ const HoloFuelDnaInterface = {
       }
     },
     getCounterparty: async ({ agentId }) => {
-      const presentCounterparty = counterparty => ({
+      const counterparty = await createZomeCall('profile/get_profile')({ agent_address: agentId })
+      if (counterparty.Err) {
```
zo-el:
```suggestion
      if (!counterparty || counterparty.Err) {
```

zo-el: This was the hot fix in #308

conductor-config-backup.toml
```diff
@@ -0,0 +1,67 @@
+bridges = []
```
zo-el: Lets just make a new branch in `holo-nixpkgs` with the right Holofuel DNA.

You can update holo-nixpkgs to : `8b122836e8074c289abe49b80755e614125fb820`

zo-el (changes requested):
Lets remove caching from HolofuelUser in the Transactions object

chaperone.json
```diff
@@ -0,0 +1,13 @@
+{
```
zo-el: We need to gitignore this file. right?

src/components/Input/Input.module.css
```diff
@@ -19,6 +19,5 @@
 
 .input-centered {
   width: 80%;
-  min-width: 280px;
```
zo-el, line 22: Can we document this change here in the PR history?

src/data-interfaces/HoloFuelDnaInterface.js
```diff
@@ -328,23 +342,22 @@ const HoloFuelDnaInterface = {
...
-      const myProfile = await HoloFuelDnaInterface.user.get()
+      // const myProfile = await HoloFuelDnaInterface.user.get()
 
       const nonActionableTransactionsWithCancelByKey = uniqueNonActionableTransactions
         .filter(tx => tx.status !== 'pending')
-        .map(tx => tx.status === STATUS.canceled ? { ...tx, canceledBy: myProfile } : { ...tx, canceledBy: null })
-
-      const presentedNonActionableTransactions = await getTxWithCounterparties(nonActionableTransactionsWithCancelByKey)
+        // nb: cancelled txs have been removed for now; keep this commented out as long as they continue being so
+        // .map(tx => tx.status === STATUS.canceled ? { ...tx, canceledBy: myProfile } : { ...tx, canceledBy: null })
```
zo-el: We can remove this because we will not be implementing cancel for this version of the DNA anyway

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
```diff
@@ -237,7 +237,8 @@ export default function CreateOfferRequest ({ history: { push } }) {
 
 export function RenderNickname ({ agentId, setCounterpartyNick, setCounterpartyFound, newMessage }) {
   const { loading, error: queryError, data: { holofuelCounterparty = {} } = {} } = useQuery(HolofuelCounterpartyQuery, {
-    variables: { agentId }
+    variables: { agentId },
+    fetchPolicy: 'network-only'
```
zo-el: Should we be using `no-cache` instead?

zo-el: Also should we just remove this Query? And we can avoid calling the new type `TransactionHolofuelUser` and instead just use `HolofuelUser`. Because we will not need `HolofuelCounterpartyQuery`

Or should we wait for a new PR?

zo-el (approved):
Please review the bellow suggestions.

Only the question about should we just remove the `HolofuelCounterpartyQuery` could be additional work that we could include in this PR

src/components/Input/Input.module.css
JettTech: This is just to adapt the input box on the Profile page to smaller mobile widths.

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
JettTech: That will be the necessary change for the ticket to remove the counterparty check, which I was planning on doing as a separate PR to keep this PR specific to the named ticket number.  Also, the functional updates to the createRequestOffer page will also be paired with UX updates.. so it will be cleaner to keep those changes apart from the focus of this PR.  

That said, the question of whether to link the functional and UX updates to the createRequestOffer page is def a product Q.  Let's follow-up with Rob on that one, and depending on his answer, we might be able to just make a new short PR which is merged into this one.

Thoughts?

zo-el: @JettTech, the only reason I bring this up now is because I think it would be easier to make those changes in the PR.

But I defer to you

chaperone.json
JettTech: Yes. Done. :+1:

src/holofuel/pages/CreateOfferRequest/CreateOfferRequest.js
JettTech: Checking with Rob.

zo-el (approved):

PR #312, opened 2020-07-24
B 02045 remove counterparty check

JettTech:
### Updates:
 - removes counterparty check when entering counterparty hash on the createOfferRequest page
 - removes counterparty check when accepting a transaction on the Inbox page
 - cleans out deprecated endpoint calls
 - updates holofuelUser to myHolofuelUser, as it is now only used as a return type for `get_my_profile`
 - updates txHolofuelUser to HolofuelUser, as this is now the general type


src/holofuel/pages/Inbox/Inbox.js
```diff
@@ -117,16 +107,16 @@ const presentTruncatedAmount = (string, number = 15) => {
...
-  const { data: { holofuelUser = {} } = {} } = useQuery(HolofuelUserQuery, { fetchPolicy: 'cache-and-network' })
+  const { data: { myHolofuelUser = {} } = {} } = useQuery(MyHolofuelUserQuery, { fetchPolicy: 'cache-and-network' })
   const { loading: ledgerLoading, data: { holofuelLedger: { balance: holofuelBalance } = {} } = {} } = useQuery(HolofuelLedgerQuery, { fetchPolicy: 'cache-and-network' })
   const { currentUser, setCurrentUser } = useCurrentUserContext()
   const { isConnected } = useConnectionContext()
 
   useEffect(() => {
-    if (!isEmpty(holofuelUser)) {
-      setCurrentUser(holofuelUser)
+    if (!isEmpty(myHolofuelUser)) {
```
zo-el, line 116: We do not use myHolofuelUser in Inbox do we?

So do we need this query  to be made in the Inbox page?

zo-el (approved):

PR #315, opened 2020-07-27
Release v0.13.0

JettTech:
Updates:
 - notification badge
 - polling cli and global var
 - connection ux updates
 - revert to mobile layout with drawer-style side-menu updates
 - remove counterparty check on login
 - correct flash message reset logic
 - counterparty nickname api changes


zo-el (changes requested):
Small things to look into before release:
- Can we reposition the model for the new layout
- Can we display a failure to create request/promise message appear on the UI (Just a simple error message for now before we get back a design from UX)
- Can we look into the timing to update the batch for a request?

zo-el (approved):

PR #316, opened 2020-07-30
B 02048 - implement new transaction toggle button designs

JettTech:


zo-el (approved):
@JettTech, I think the Send and Receive text shown bellow should be made bold.
Its to light on the screen.
Screenshot from 2020-07-30 11-19-06

Else everything looks good

PR #322, opened 2020-08-03
B-01811 - HP Admin can get the number of zome calls made to a hosted hApp

JettTech:
### Updates;
 - plugs in zome call endpoint


src/mock-dnas/hpos.js
```diff
@@ -55,7 +55,14 @@ const data = {
       hosted_happs: [
         {
           'happ-title': 'HoloFuel',
-          number_instances: 17
+          number_instances: 17,
+          stats: {
+            traffic: {
+              start_date: '2020-07-27T21:27:25Z',
+              total_zome_calls: 100,
+              values: [10, 40, 20, 30]
```
robbiecarlton: Only change I would make is to not mock any data we're not using. It just creates debt. If we need this extra data later we can just add it to the mocks at that time

robbiecarlton (changes requested):
One small change, otherwise good to go.

robbiecarlton (approved):

zo-el (approved):

JettTech:
nb: Tests fail on ci due to memory, but pass locally.

PR #325, opened 2020-08-03
B-01990 - Add ssh access control to Settings

robbiecarlton:


src/pages/Settings/Settings.js
```diff
@@ -74,6 +75,27 @@ export function Settings () {
 
   const title = (settings.hostName ? `${settings.hostName}'s` : 'Your') + ' HoloPort'
 
+  const [sshAccess, setSshAccess] = useState(settings.sshAccess)
+  const saveSshAccess = () => {
+    console.log('sshAccess', sshAccess)
```
JettTech: remove log

src/pages/Settings/Settings.js
```diff
@@ -165,4 +195,19 @@ function VersionUpdateButton ({ updateVersion }) {
   </Button>
 }
 
+export function SettingsFormInput ({
```
JettTech, line 204: As this is only being used for the checkbox, can we update the name to reflect that?

JettTech (changes requested):
One question, one request for change:
 1. I thought the original designs were for a toggle button instead of a checkbox.  It does look like the ssh button was removed from figma designs for reference, but do we have new design decisions for this? 
2. Also, it appears that when I click on the button on, the data sends ssh as false, and when the checkbox is not checked, the ssh data is sent as true.

image

JettTech:
Also, per our sync this morning, it seems there might be new endpoints to review. Let's follow-up on this together.

PR #326, opened 2020-08-03
D 01114 - update feedback message and flow upon unsuccessful attempt to create new transaction

JettTech:
### Updates:
 - change copy per po decisions
 - update reroute to inbox

### DIFF
--- a/src/models/Transaction.js
+++ b/src/models/Transaction.js
@@ -15,6 +15,8 @@ export const DIRECTION = {
   outgoing: 'outgoing'
 }
 
+export const FAILED_TRANSACTION_MESSAGE = 'The recipient for the transaction could not be found.  The transaction will appear as declined'
+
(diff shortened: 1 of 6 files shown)


zo-el (approved):
Works. Lets get this in

src/models/Transaction.js
```diff
@@ -15,6 +15,8 @@ export const DIRECTION = {
   outgoing: 'outgoing'
 }
 
+export const FAILED_TRANSACTION_MESSAGE = 'The recipient for the transaction could not be found.  The transaction will appear as declined'
```
zippy, line 18: Are we ending these messages with punctuation?  If so needs a period.

JettTech: Currently, we haven't been, but let's circle back with Rob and C on this one, if it's an interest.  Otherwise, we're at least remaining consistent. :+1:

PR #331, opened 2020-08-06
B-01938 Desktop Layout

robbiecarlton:


src/components/Header/Header.test.js
```diff
@@ -9,14 +9,14 @@ it('should render the title and a menu icon', async () => {
   const props = {
     title: 'the title',
     settings: {
-      hostName: 'hostname'
+      deviceName: 'devicename'
```
JettTech, line 12: :+1:

JettTech (changes requested):
Overall looks good.  I did notice that the content doesn't fit on the page when rendering the traditional tablet viewport width (see the screenshot below for context). Let's update the responsive styles so that nothing is ever off-page (..and maybe set a min-width for the side menu) and then this ticket is good to merge. 

image

JettTech (approved):
lgtm

PR #332, opened 2020-08-11
D-01126 - Hot Fix - Updated id to agentAddress

JettTech:
### Updates: 
 - restores copy functionality to agentHash

### DIFF
--- a/src/components/CopyAgentId/CopyAgentId.js
+++ b/src/components/CopyAgentId/CopyAgentId.js
@@ -10,17 +10,16 @@ export default function CopyAgentId ({
   className,
   children
 }) {
-
   let happName, hash, nickname, messageText
 
   if (hpAdmin) {
     ({ id: hash, nickname } = agent)
-    happName = 'HP Admin' 
+    happName = 'HP Admin'
   } else {
     ({ agentAddress: hash, nickname } = agent)
-    happName = 'HoloFuel' 
+    happName = 'HoloFuel'
   }
-  
+
   if (isMe) {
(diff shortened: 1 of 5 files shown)


robbiecarlton (approved):
The code as it stands has redundancy because we have two different `CopyAgentId` components, one for hp-admin and one for holofuel, and yet in the hp-admin one, we are also checking which app we're in. It would be better if we picked one approach or the other. The code works so I'm good to merge it but I'm also going to work on another fix.

robbiecarlton:
@JettTech - Here's a PR into this branch that just removes the holofuel case in the hp-admin `CopyAgentId` component. I went through the app and as far as I can tell we weren't using that case anywhere.

https://github.com/Holo-Host/hp-admin/pull/338/files

LMK what you think

robbiecarlton (approved):

JettTech:
Yes, this is pattern is redundant, and I was always against the decision made last year to include this pattern of redundancy.  Even though I agree with this clean-up work and am still am not in favor of the level of redundancy within this codebase, I also don't know how much effort is worth allocating toward cleaning up a version we'll soon be leaving.  - I think we both agree that the next version is already off to a much cleaner start; so let's just see that we don't perpetuate any redundant pattern in the next version (which obviously shouldn't be too much of an issue since we're not combining two codebases into one repo).

Nonetheless, I merged the clean-up work in and updated the tests accordingly.

As changes have been implemented since your last review, please review once more before merging.

robbiecarlton:
Approved

PR #334, opened 2020-08-11
D-01128 Github Pages routing issue

robbiecarlton:


src/holofuel/HFRouter.js
```diff
@@ -26,7 +24,7 @@ export default function HFRouter () {
     <HFRoute path={`${root}/offer-request`} exact component={CreateOfferRequest} />
     <HFRoute path={`${root}/profile`} exact component={Profile} />
     <HFRoute path={`${root}`} exact component={() => <Redirect to={`${root}/`} />} />
```
JettTech, line 26:
```suggestion
    <HFRoute path={`${root}`} exact component={() => <Redirect to={`${root}/inbox`} />} />
```

JettTech: Shouldn't use the same redirect path here?

JettTech (approved):
This solution works, but let's cycle back on the long term decision for servers.

Merge it! :shipit:

PR #337, opened 2020-08-31
HposInterface.js: fix heders for delete http verb

peeech:
By best guess at how to fix DELETE 401 :-)


robbiecarlton (approved):

robbiecarlton:
Nice catch

